PR #9, opened 2024-12-17
chore: Fix issues flagged by SonarCloud to please our AI overlords

arturbalabanov:
* [x] Get the code coverage up to 95%
* [x] Improve the coverage config to calculate it correctly (it was missing some async calls before)
* [x] Remove commented out code
* [x] Skip the sonarcloud steps from CI if run with [act](https://github.com/nektos/act)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,6 @@ dependencies = [
 [dependency-groups]
 dev = [
     "bandit>=1.8.0,<2.0",
-    "gevent>=24.11.1",
     "httpx>=0.27.2,<1.0",
     "ipython>=8.30.0,<9.0",
     "pre-commit>=4.0.1,<5.0",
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -84,6 +85,9 @@ cache_dir = ".cache/pytest"
 branch = true
 source = ["app"]
 relative_files = true
+# gevent is needed for fastapi projects with async/await calls, otherwise the coverage misses them
+# ref: https://stackoverflow.com/a/76033894
+concurrency = ["gevent"]
```
ffaraone, line 90: You can use greenlet here and avoid installing gevent.
Looks like SQLAlchemy uses greenlets to add asyncio support.

ffaraone (changes requested):

pyproject.toml
antoniodimariano: I have a question: 
I'd like to understand why we need Gevent here. 
Gevent is a library for green threads. It is typically used in projects relying on synchronous code where concurrency is desired. FastAPI already embraces asyncio, making Gevent redundant, in my opinion.
How could it directly impact test coverage?

ffaraone: In fact is not related to fastapi. is related to sqlalchemy. sqlalchemy already depends on greenlet library so the only thing that is needed to collect the coverage the right way is to specify to coverage the concurrency libraries used in the project (see [https://coverage.readthedocs.io/en/7.6.9/config.html#run-concurrency](https://coverage.readthedocs.io/en/7.6.9/config.html#run-concurrency). Probably the right solution here is to specify both greenlet and thread:

```
concurrency = ["thread", "greenlet"]
```

arturbalabanov: Thank you for this guys, learning new things every day :) Changed to `["thread", "greenlet"]` and updated the comment to be more accurate, also included these links

ffaraone (approved):
lgtm

PR #10, opened 2024-12-18
feat: Add status to Entitlements

arturbalabanov:
Seems like the entitlements requirements page includes status - either a new update or I missed it initially, so adding it now

Didn't include the other fields for now to keep it simple. Will add that in future PRs when we add `/v1/entitlements/{id}/activate` and `/v1/entitlements/{id}/terminate`

Also fix a small bug where the server returns 500 instead of 422 when passing invalid UUID in the URL


app/routers/entitlements.py
```diff
@@ -27,6 +29,6 @@ async def create_entitlement(data: EntitlementCreate, session: DBSession):
 
 
 @router.patch("/{id}", response_model=EntitlementRead)
```
ffaraone, line 31: Apologize I miss it in the previous PR, there is no patch for entitlements, please check the TDR article. Thank you!

ffaraone (changes requested):

ffaraone:
> Seems like the entitlements requirements page includes status - either a new update or I missed it initially, so adding it now
> 
> Didn't include the other fields for now to keep it simple. Will add that in future PRs when we add `/v1/entitlements/{id}/activate` and `/v1/entitlements/{id}/terminate`
> 
> Also fix a small bug where the server returns 500 instead of 422 when passing invalid UUID in the URL

The transition from new to active will not be done via api using the `activate` action, please check the TDR, thank you

app/routers/entitlements.py
arturbalabanov: Removed the endpoint and its tests :)

ffaraone (approved):

PR #12, opened 2024-12-24
MPT-5735 split api into namespaces (billing, account)

ffaraone:


ffaraone:
Namespaces must be discussed, I will close such PR.

PR #13, opened 2024-12-27
MPT-4905 add create organization endpoint

ffaraone:


app/constants.py
```diff
@@ -0,0 +1,4 @@
+API_MODIFIER_JWT_ALGORITHM = "HS256"
+API_MODIFIER_JWT_ISSUER = "SWO"
+API_MODIFIER_JWT_AUDIENCE = "modifier"
+API_MODIFIER_JWT_EXPIRE_AFTER_SECONDS = 300
```
antoniodimariano, line 4: If we moved `API_MODIFIER_JWT_EXPIRE_AFTER_SECONDS` to the .env, wouldn't it be easier to change its value?

antoniodimariano (approved):

app/constants.py
ffaraone: I don't think it's subject to be changed often, really.

PR #14, opened 2024-12-27
MPT-4908 get organization details

ffaraone:


tests/conftest.py
```diff
@@ -102,12 +102,15 @@ def organization_factory(
...
-        name: str | None = None, external_id: str | None = None
+        name: str | None = None,
+        external_id: str | None = None,
+        organization_id: str | None = None,
     ) -> Organization:
         return await organizations_handler.create(
             Organization(
                 name=name or faker.company(),
                 external_id=external_id or "ACC-1234-5678",
+                organization_id=organization_id or None,
```
antoniodimariano, line 113: We probably live with just `organization_id` here

ffaraone: Good catch!

PR #15, opened 2024-12-30
MPT-5730 add extension (System) user

ffaraone:


app/db/handlers.py
```diff
@@ -40,41 +36,41 @@ async def create(self, data: BaseModel) -> T:
...
+            setattr(obj, key, value)
+        await self.session.commit()
+        await self.session.refresh(obj)
+        return obj
 
-    async def fetch_page(self, limit: int = 50, offset: int = 0) -> Sequence[T]:
-        results = await self.session.exec(select(self.model_cls).offset(offset).limit(limit))
-        return results.all()
+    async def fetch_page(self, limit: int = 50, offset: int = 0) -> Sequence[ModelT]:
+        results = await self.session.execute(select(self.model_cls).offset(offset).limit(limit))
```
antoniodimariano, line 56: What if a negative value is passed in limit and/or offset? I know it's probably not the case,but 

`if limit < 0 or offset < 0:
            raise ValueError("Limit and offset must be non-negative integers.")
`
Also, there is no way to catch an exception that the query might raise

app/routers/organizations.py
```diff
@@ -43,8 +48,13 @@ async def create_organization(data: OrganizationCreate, organization_repo: Organ
...
-            organization.organization_id = ffc_organization["id"]
-            return await organization_repo.update(organization.id, organization)
+            db_organization = await organization_repo.update(
+                db_organization.id,
+                {
+                    "organization_id": ffc_organization["id"],
+                },
+            )
+            return OrganizationRead.from_orm(db_organization)
     except httpx.HTTPStatusError as e:
```
antoniodimariano, line 58: I would consider an alternative way to express the exceptions, using just one try/except block and inserting all the exception catching at the end.  What do you think?

app/routers/entitlements.py
```diff
@@ -3,30 +3,48 @@
...
+    data: EntitlementCreate,
+    entitlement_repo: EntitlementRepository,
+    system: CurrentSystem,
+):
+    entitlement = data.to_orm()
+    entitlement.created_by = system
+    entitlement.updated_by = system
+
+    db_entitlement = await entitlement_repo.create(entitlement)
+    return EntitlementRead.from_orm(db_entitlement)
```
antoniodimariano, line 34: I would add a generic `except Exception as error` block to manage unexpected errors

antoniodimariano (approved):

app/routers/organizations.py
ffaraone: In this case we are returning two different errors with two different http statuses, so I think it's better to leave it as is to distinguish between errors received from the modifier from errors of the operations api itself.

app/db/db.py
```diff
@@ -16,7 +19,7 @@
 )
 
 
-async def get_db_session() -> AsyncIterator[AsyncSession]:
+async def get_db_session() -> AsyncGenerator[AsyncSession]:
     async_session = async_sessionmaker(bind=db_engine, class_=AsyncSession, expire_on_commit=False)
```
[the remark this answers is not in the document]
ffaraone: Yeah, this must be changed, we also need to change the way we control transactions so I will open a specific task to fix such things.

app/routers/entitlements.py
ffaraone: Probably we should implement a custom fastapi exception handler to handle also such unexpected errors. I will open a task for that.

app/db/handlers.py
ffaraone: I will do in my next task.

ffaraone:
I created the following to handle @antoniodimariano  feedback:

- MPT-5773
- MPT-5774
- MPT-5776

PR #20, opened 2025-01-03
refactor: extract API calls to the Optscale API, Optscale Auth API and the API Modifier into separate clients

arturbalabanov:
Please let me know what you think of this design. I intentionally kept the test changes as minimal as possible to hear your thoughts on the approach before committing more time. However, with the current design, the tests can be simplified a lot and we can easily create reusable mock objects with whatever information we need.

Feel free to review commit by commit as there are two other changes I bundled here, again let me know about your thoughts on these:

1. 65caa3acfcbc74f023f3c20798f585833c6855f0 -- allows for the fastapi lifespan to be executed when using the async testing client
2.  acbd949892e4d28ad61757f9b2a8395b5ab6e981 -- integrates the [`svcs`](https://svcs.hynek.me/en/stable/index.html) package for managing the clients and reduces boilerplate for creating dependancies for them. I think it can be used for many other things but I'm open to your thoughts as I haven't used it before. I found [this video](https://www.youtube.com/watch?v=d1elMD9WgpA) by the author to give a great explanation about the package's architecture and the problems it's solving

### DIFF
--- a/app/api_clients/__init__.py
+++ b/app/api_clients/__init__.py
@@ -1,0 +1,8 @@
+from app.api_clients.api_modifier import *  # noqa: F401, F403
+from app.api_clients.base import BaseAPIClient  # noqa: F401
+from app.api_clients.optscale import *  # noqa: F401, F403
+from app.api_clients.optscale_auth import *  # noqa: F401, F403
+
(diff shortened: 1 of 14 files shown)


ffaraone (approved):
👍 Fine for me

PR #23, opened 2025-01-07
MPT-5859 fix create organization when partially created (db only)

ffaraone:


app/db/handlers.py
```diff
@@ -49,6 +49,25 @@ async def get(self, id: UUID | str) -> M:
                 f"Failed to get {self.model_cls.__name__} with ID `{str(id)}`: {e}"
             ) from e
 
+    async def get_or_create(self, defaults: dict[str, Any], **filters: Any) -> tuple[M, bool]:
```
arturbalabanov, line 52: I think `defaults` should be a keyword-only argument to avoid usage confusion. It's also not strictly necessary in all cases, so we can make it optional too. Similarly, I think most of the time we want to fetch an object by its ID, so we can use that as the first argument instead, i.e. the function signature would be:

```python
async def get_or_create(self, id: str | UUID, *, defaults: dict[str, Any], **filters: Any) -> tuple[M, bool]
```

What do you think?

arturbalabanov (approved):

app/db/handlers.py
ffaraone: Agree 100%

PR #25, opened 2025-01-08
feat: Add terminate entitlement endpoint (MPT-4903)

arturbalabanov:


app/main.py
```diff
@@ -40,10 +41,11 @@ async def lifespan(app: FastAPI, registry: svcs.Registry):
     root_path="/v1",
     debug=settings.debug,
     lifespan=lifespan,
+    dependencies=[Depends(get_current_system)],
```
ffaraone, line 44: I'm not sure this optimization will survive the introduction of roles and operators 😢. Anyway we can leave it as you changed and we will see.

app/routers/entitlements.py
```diff
@@ -35,17 +36,51 @@ async def create_entitlement(
...
+async def get_entitlement_by_id(
+    entitlement: Annotated[Entitlement, Depends(fetch_entitlement_or_404)],
+):
+    return from_orm(EntitlementRead, entitlement)
+
+
+@router.post("/{id}/terminate", response_model=EntitlementRead)
+async def terminate_entitlement(
+    entitlement: Annotated[Entitlement, Depends(fetch_entitlement_or_404)],
+    session: DBSession,
```
ffaraone, line 61: We should use the entitlement repository to transition to terminated right ? If we use the db session the db access layer has no reason to exist.

ffaraone (changes requested):

app/routers/entitlements.py
arturbalabanov: Moved to the repository and did some refactoring to get the current system from anywhere in the app via context managers. This allows it to not be passed around to the db layer (needed in order to set `terminated_by` and `updated_by`) as that would be done for basically any api call which does a write db operation. Also don't like the idea of having to manually set these fields (this can still be done if we want to overwrite the current behaviour) as it can be easily forgotten.

What do you think? It's a bigger change than I initially anticipated and may quickly become obsolete with the introduction of the roles and operators but I think the general principle of using context vars for these kind of things will still be relevant

ffaraone (approved):
It starts looking really awesome! 🥇

PR #27, opened 2025-01-16
feat: Add endpoints to list an organization's cloud accounts (MPT-4914, MPT-4915)

arturbalabanov:


app/api_clients/base.py
```diff
@@ -41,9 +27,29 @@ def __init__(self, message: str):
         super().__init__(f"{self.client_name} API client error: {message}")
 
 
+class JWTTokenAuth(httpx.Auth):
```
ffaraone, line 30: I would suggest to be a bit more explicit here.
This class handle the specific authentication needed by the modifier not a generic authentication based on JWT.

app/api_clients/base.py
```diff
@@ -41,9 +27,29 @@ def __init__(self, message: str):
...
+    def auth_flow(self, request: httpx.Request) -> Generator[httpx.Request, httpx.Response, None]:
+        # NOTE: Needs to be re-generated for each request as it exipres after a certain time
+        jwt_token = get_api_modifier_jwt_token()
+
+        request.headers["Authorization"] = f"Bearer {jwt_token}"
+
+        yield request
+
+
+class ClusterSecretAuth(httpx.Auth):
```
ffaraone, line 40: Again, even if the concept of ClusterSecret is something specific of OptScale I would prefer to be more specific here.

app/routers/organizations.py
```diff
@@ -80,3 +82,91 @@ async def get_organization_by_id(id: UUID, organization_repo: OrganizationReposi
             status_code=status.HTTP_404_NOT_FOUND,
             detail=str(e),
         ) from e
+
+
+@router.get("/{id}/cloud-accounts", response_model=list[CloudAccountRead])
+async def get_cloud_accounts_by_organization_id(
+    id: UUID,
```
ffaraone, line 89: I think we should use the same approach as we used for entitlement right ? like `get_organization_or_404`

ffaraone (approved):

app/api_clients/base.py
arturbalabanov: You're right, changed the names of both auth classes to be more descriptive

app/routers/organizations.py
arturbalabanov: Good spot, added the dependancy :)

PR #29, opened 2025-01-20
MPT-4913 list organization's users

ffaraone:


app/routers/organizations.py
```diff
@@ -160,3 +160,39 @@ async def get_cloud_account_by_id(
...
+
+    users = response.json()["employees"]
+
+    return [
+        UserRead(
+            id=user["id"],
+            email=user["user_email"],
+            display_name=user["user_display_name"],
+            created_at=user["created_at"],
+            last_login=user["last_login"],
```
arturbalabanov, line 194: These two fields are integers in the Optscale's response, so we need to use `datetime.fromtimestamp(..., datetime.timezone.utc)` here instead.

Also, according to the docs none of these fields are nullable but they are on our side. `created_at` will never be null, right? As for `last_login` I imagine the Optscale API will return 0 when the user has never logged in and we need to handle that

app/schemas.py
```diff
@@ -112,6 +112,9 @@ class OrganizationUpdate(OrganizationBase):
 class UserBase(BaseSchema):
     email: Annotated[str, Field(max_length=255)]
     display_name: Annotated[str, Field(max_length=255)]
+    created_at: datetime.datetime | None = None
+    last_login: datetime.datetime | None = None
+    roles_count: int | None = None
```
arturbalabanov, line 117: Is there a case when `roles_count` would be `None`? What would that mean (i.e. what's the difference between that and 0?

arturbalabanov: Actually I just saw that was the base model, so that makes sense for the create model. But we need to overwrite it for `UserRead`

app/routers/organizations.py
ffaraone: Looks like pydantic does enough black magic like it understand you are creating the object with the timestamp. They allow None because this model is shared between the response of create user where we get only id, email and display_name and the list of users within an organization. So the three new attribute are by default None.
Another approach could be to distinguish between the response of the create user and the response of list users but I don't think there is a need to do that.

app/schemas.py
ffaraone: Read my previous comment about `UserRead` please.

app/routers/organizations.py
arturbalabanov: I see, thanks for the explanation! That all makes sense, no need for any changes then

> Another approach could be to distinguish between the response of the create user and the response of list users but I don't think there is a need to do that.

I suppose that's technically more correct and we may want to do it in the future but you're right, it would be overcomplicating it for now

arturbalabanov (approved):

PR #36, opened 2025-01-28
feat: Add tests and play arround with the DB handlers to allow for session.begin() transactions (MPT-5773)

arturbalabanov:


arturbalabanov:
Bye bye demo PR -- you had a short but very meaningful life and you'll always be remembered in our hearts ❤️

PR #37, opened 2025-01-28
MPT-5773 Transaction management

ffaraone:
Let's discuss this proposal.

### DIFF
--- a/app/db/db.py
+++ b/app/db/db.py
@@ -12,6 +12,11 @@ from sqlalchemy.ext.asyncio import (
 
 from app import settings
 
+
+class AsyncTxSession(AsyncSession):
+    pass
+
+
 db_engine = create_async_engine(
     str(settings.postgres_async_url),
     echo=settings.debug,
@@ -23,6 +28,17 @@ async def get_db_session() -> AsyncGenerator[AsyncSession]:
     async_session = async_sessionmaker(bind=db_engine, class_=AsyncSession, expire_on_commit=False)
     async with async_session() as session:
         yield session
+
+
+@asynccontextmanager
+async def get_tx_db_session() -> AsyncGenerator[AsyncSession]:
+    async_session = async_sessionmaker(
+        bind=db_engine, class_=AsyncTxSession, expire_on_commit=False
+    )
+    async with async_session() as session:
+        async with session.begin() as transaction:
+            yield session
+            await transaction.commit()
 
 
 DBSession = Annotated[AsyncSession, Depends(get_db_session)]
--- a/app/db/handlers.py
+++ b/app/db/handlers.py
@@ -8,6 +8,7 @@ from sqlalchemy.exc import DBAPIError, IntegrityError
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from app.auth import current_system
+from app.db.db import AsyncTxSession
 from app.db.models import AuditableMixin, Entitlement, Organization, System
 from app.db.models import Base as BaseModel
 from app.enums import EntitlementStatus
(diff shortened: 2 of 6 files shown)


app/db/handlers.py
```diff
@@ -124,7 +131,10 @@ async def terminate(self, entitlement: Entitlement) -> Entitlement:
         entitlement.updated_by = current_system.get()
         entitlement.terminated_at = datetime.now(UTC)
 
-        await self.session.commit()
+        if self.commit:
+            await self.session.commit()
+        else:
+            await self.session.flush()
```
antoniodimariano, line 137: I would avoid the repetition of 
```
if self.commit:
            await self.session.commit()
        else:
            await self.session.flush()
 ```
 by creating a function like 
 
 ```
 async def _commit_or_flush(self)->None:
      if self.commit:
           await self.session.commit()
      else:
          await self.session.flush()
 ```

File: app/db/db.py [hunk 2]
```diff
@@ -25,6 +30,17 @@ async def get_db_session() -> AsyncGenerator[AsyncSession]:
...
 
+@asynccontextmanager
+async def get_tx_db_session() -> AsyncGenerator[AsyncSession]:
+    async_session = async_sessionmaker(
+        bind=db_engine, class_=AsyncTxSession, expire_on_commit=False
+    )
+    async with async_session() as session:
+        async with session.begin() as transaction:
+            yield session
+            await transaction.commit()
```
arturbalabanov, line 41: A few notes on this context manager:

1. `transaction.commit` is not necessary here as that's what `session.begin()` already handles (including rolling back on exception etc.) https://docs.sqlalchemy.org/en/20/orm/session_api.html
2. I think it would be best to yield the transaction itself (the type is `AsyncSessionTransaction`), what do you think? The transaction can be used the same way as the session but it's more explicit
3. I think `AsyncTxSession` is not needed here, what's its purpose? It can also cause confusion with the already existing `AsyncSessionTransaction` from sqlalchemy
4. You can do minor refactoring to reuse the common logic from `get_db_session`

I don't feel strongly about any of these though, so feel free to resolve as you see fit. In any case the transaction and session model we use will evolve as our application needs get more complex, so don't want to discourage refactoring here

app/db/handlers.py
```diff
@@ -28,6 +29,7 @@ class ConstraintViolationError(DatabaseError):
 class ModelHandler[M: BaseModel]:
     def __init__(self, session: AsyncSession) -> None:
         self.session = session
+        self.commit = not isinstance(self.session, AsyncTxSession)
```
arturbalabanov, line 32: Ah, that explains one of my previous questions. I think a better way to handle this would be to either check for instance of the `AsyncSessionTransaction` class or to use `session.in_transaction`. Also, don't we want to overwrite this explicitly?

app/db/handlers.py
```diff
@@ -115,18 +115,24 @@ async def count(self) -> int:
...
-
-        return entitlement
+        return await self.update(
+            entitlement.id,
+            {
+                "status": EntitlementStatus.TERMINATED,
+                "terminated_at": datetime.now(UTC),
+                "terminated_by": current_system.get(),
+            },
+        )
```
arturbalabanov, line 135: Good change! :)

tests/utils.py
```diff
@@ -18,3 +24,95 @@ def assert_json_contains_model(json: dict[str, Any], expected_schema: BaseReadSc
             raise AssertionError(f"{expected_schema} has no attribute {key}")
 
         assert expected_dict[key] == actual_value
+
+
+class SQLAlchemyCapturer:
```
arturbalabanov, line 29: Amazing! I was thinking in very similar terms, so building on top of this will be very natural. Great minds think alike! 🤣

arturbalabanov (approved):
Approving this, so that I don't block this being merged. Left a few comments, feel free to resolve as you see fit and it's something that will naturally evolve anyway

PR #40, opened 2025-02-03
MPT-6689 add support for user authentication

ffaraone:


app/db/handlers.py
```diff
@@ -66,12 +82,23 @@ async def create(self, obj: M) -> M:
...
 
-    async def get(self, id: str) -> M:
+    async def get(
+        self, id: str, extra_conditions: list[ColumnExpressionArgument] | None = None
+    ) -> M:
+        extra_conditions = extra_conditions or self.default_extra_conditions
         try:
-            result = await self.session.get(self.model_cls, id, options=self.default_options)
-            if result is None:
+            query = select(self.model_cls).where(
```
antoniodimariano, line 90: why not 
```
if result is not None:
   return result 
```
so if we are lucky, we return early

ffaraone: discussed offline

antoniodimariano (approved):

PR #41, opened 2025-02-05
feat: Create a pytest fixture to allow us to assert for sql expressions sqlalchemy generates (MPT-6619)

arturbalabanov:


ffaraone (commented):
Looks clean and clear, kudos!
I have a question: does it support nested transactions ?

Anyway there is something that's brain damaging  to me: tests that test a test utility :scream:!

I would suggest to put this PR on-hold and as soon as we have some time we can move it to its own repo and publish it to pypi. Then we can use it in this project.

arturbalabanov:
> Looks clean and clear, kudos! I have a question: does it support nested transactions ?

Thanks! Not right now but it can be very easily extendible to do that -- just adding two more hooks. I was debating about doing it but wanted to finish it more quickly

> Anyway there is something that's brain damaging to me: tests that test a test utility 😱!

I knooow, same! As you said yourself, this should be extracted into a seperate package, should be very easy as I made sure to minimise the internal dependancies

> I would suggest to put this PR on-hold and as soon as we have some time we can move it to its own repo and publish it to pypi. Then we can use it in this project.

Not a bad idea though it can still be useful to have it in the project until then as we can use it and even learn about how we can improve it's design. Anyway, happy with your suggestion too (will go ahead and resolve the jira issue though, so that we don't skew the metrics)

ffaraone:
I will close since we published [pytest-capsqlalchemy](https://github.com/softwareone-platform/pytest-capsqlalchemy).

PR #42, opened 2025-02-06
feat: Add get system by id endpoint (MPT-6621)

arturbalabanov:


app/routers/systems.py
```diff
@@ -1,45 +1,62 @@
...
+from app.db.models import System
+from app.dependencies import CurrentAuthContext, SystemId, SystemRepository
+from app.schemas import SystemCreate, SystemRead, SystemUpdate, from_orm
 
 router = APIRouter()
 
 
+async def fetch_system_or_404(id: SystemId, system_repo: SystemRepository) -> System:
+    try:
+        return await system_repo.get(id=id)
```
ffaraone, line 17: We should put the following logic here:
```suggestion
        extra_conditions = []
        if auth_context.account.type == AccountType.AFFILIATE:
            extra_conditions.append(System.owner == auth_context.account)
        return await system_repo.get(id=id, *extra_conditions)
```

app/routers/systems.py
```diff
@@ -1,45 +1,62 @@
...
     pass
 
 
 @router.get("/{id}", response_model=SystemRead)
-async def get_system_by_id(id: UUID):
-    pass
+async def get_system_by_id(
+    system: Annotated[System, Depends(fetch_system_or_404)], auth_ctx: CurrentAuthContext
+):
+    if system.owner_id != auth_ctx.account.id:
```
ffaraone, line 39: Please check my comment on `fetch_system_or_404`

tests/conftest.py
```diff
@@ -109,7 +109,7 @@ async def api_client(fastapi_app: FastAPI, app_lifespan_manager: LifespanManager
 
 
 @pytest.fixture
-def account_factory(faker: Faker, db_session: AsyncSession) -> ModelFactory[Account]:
+async def account_factory(faker: Faker, db_session: AsyncSession) -> ModelFactory[Account]:
```
ffaraone, line 112: why you want to make such factories async ?

app/routers/systems.py
arturbalabanov: Added, thanks

tests/conftest.py
arturbalabanov: Good catch, we don't need to do that, reverted it. Leftover from debugging the tests

app/db/handlers.py
```diff
@@ -85,7 +85,11 @@ async def create(self, obj: M) -> M:
     async def get(
         self, id: str, extra_conditions: list[ColumnExpressionArgument] | None = None
     ) -> M:
-        extra_conditions = extra_conditions or self.default_extra_conditions
+        extra_conditions = extra_conditions or []
```
ffaraone, line 88: I think this changes must be discussed first. The original idea (probably wrong but intentional) was that the `extra_conditions` overrides the default ones not extend them.

ffaraone (changes requested):
Probably we should be explicit

app/db/handlers.py
ffaraone: Finally I removed such default_extra_conditions, I think it's not clear. Please rebase your PR on latest main. Thank you!

app/auth/auth.py
```diff
@@ -59,10 +60,14 @@ async def get_authentication_context(
...
-            system = await system_handler.get(
-                actor_id,
-                [System.status == SystemStatus.ACTIVE],
-            )
+            try:
+                system = await system_handler.get(
+                    actor_id,
+                    [System.status == SystemStatus.ACTIVE],
+                )
+            except NotFoundError as exc:
```
ffaraone, line 68: I don't think we need this try/except block since `NotFoundError` inherits from `DatabaseError` that is handled by outer try/except block. Removing it will reduce a bit the cognitive complexity.

tests/conftest.py
```diff
@@ -194,6 +194,7 @@ async def _system(
         owner: Account | None = None,
         status: SystemStatus = SystemStatus.ACTIVE,
     ) -> System:
+        owner = owner or await account_factory()
```
ffaraone, line 197: Either remove this line either change the line 202 please.

ffaraone (approved):

PR #53, opened 2025-02-12
feat: Add endpoint for disabling a system (MPT-6529)

arturbalabanov:


app/routers/systems.py
```diff
@@ -84,8 +84,21 @@ async def delete_system_by_id(id: SystemId):  # pragma: no cover
...
+    if system.status != SystemStatus.ACTIVE:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=(
+                f"System's status is '{system.status._value_}'; "
+                "only active systems can be disabled"
+            ),
+        )
+
+    system = await system_repo.update(system, {"status": SystemStatus.DISABLED})
```
ffaraone, line 100: We should check that a System cannot disable itself.

ffaraone (changes requested):

ffaraone (approved):

PR #60, opened 2025-02-14
Mpt 6517 account update endpoint

antoniodimariano:
Add PUT /accounts/{id}
Add Tests

### DIFF
--- a/app/routers/accounts.py
+++ b/app/routers/accounts.py
@@ -35,33 +35,36 @@ async def persist_data_and_format_response(account_repo, data):
     return from_orm(AccountRead, db_account)
 
 
-async def update_data_and_format_response(id,account_repo, data):
+async def update_data_and_format_response(
+    id: str, account_repo: AccountRepository, data: AccountUpdate
+):
     """
     It updates the given data to the Account model and return back
     the Pydantic schema
 
     account_repo: an ORM model instance of AccountRepository
-    data: the data to persist
+    data: the data to update
     Return: AccountUpdate Pydantic Model
     """
-    print("data:", type(data))
     to_update = data.model_dump(exclude_none=True)
-    print("to_update:",to_update)
     if not to_update:
         raise HTTPException(
             status_code=status.HTTP_400_BAD_REQUEST,
             detail="You can't update whatever you want.",
         )
-    db_account = await account_repo.update(id=id, data=data.model_dump(exclude_none=True))
-    print("db_account", db_account.name)
-    from_orm_ =  from_orm(AccountRead, db_account)
-    print("from orm",from_orm_)
+    db_account = await account_repo.update(id, data=data.model_dump(exclude_none=True))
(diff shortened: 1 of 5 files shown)


app/routers/accounts.py
```diff
@@ -35,6 +35,49 @@ async def persist_data_and_format_response(account_repo, data):
...
+    """
+    to_update = data.model_dump(exclude_none=True)
+    if not to_update:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="You can't update whatever you want.",
+        )
+    db_account = await account_repo.update(id, data=data.model_dump(exclude_none=True))
+    from_orm_ = from_orm(AccountRead, db_account)
+    return from_orm_
```
ffaraone, line 57:
```suggestion
db_account = await account_repo.update(id, to_update)
return from_orm(AccountRead, db_account)
```

app/routers/accounts.py
```diff
@@ -35,6 +35,49 @@ async def persist_data_and_format_response(account_repo, data):
...
+    2. The account status must be ACTIVE.
+
+    A HTTPException with status 400 will be raised if at least one condition is not met.
+    """
+    if account.type != AccountType.AFFILIATE:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="You cannot update an Account of type Operations.",
+        )
+    if account.status == AccountStatus.DELETED or account.status == AccountStatus.DISABLED:
```
ffaraone, line 74: Why we deny to update a disabled account ?

app/routers/accounts.py
```diff
@@ -116,16 +159,58 @@ async def get_accounts(account_repo: AccountRepository):
...
+    This Endpoint updates an Affiliate Account.
+
+    The following conditions must be verified before proceeding with the operation of updating
+    an account.
+    1. The Account type must be OPERATIONS, otherwise a 403 error will be returned
+    2. The Account status must be Active
+    3. Only Accounts classified as of type “Affiliate” can be updated.
+    4. Only the name and the external_id of the account can be modified.
+
+    The newly created account's status will be assigned as ACTIVE
```
ffaraone, line 182: 1. This operation can be performed only in the context of an Operations Account otherwise 403 will be returned.
2. Deleted Accounts cannot be updated.

The sentence

"The newly created account's status will be assigned as ACTIVE" looks copy/pasted from the create endpoint.

app/routers/accounts.py
```diff
@@ -116,16 +159,58 @@ async def get_accounts(account_repo: AccountRepository):
...
+
+    Raises:
+        - HTTPException with status 403 if the check (1) or (3)fails
```
ffaraone, line 185: in case of 3) the status is 400 right ?

ffaraone (changes requested):

PR #61, opened 2025-02-14
MPT-6903 redeem entitlements command

ffaraone:


app/commands/redeem_entitlements.py
```diff
@@ -0,0 +1,114 @@
...
+        f"{ctx.obj.opt_api_base_url}/organizations/{organization_id}/cloud_accounts",
+        params={
+            "details": "true",
+        },
+        headers={
+            "Secret": ctx.obj.opt_cluster_secret,
+        },
+    )
+    response.raise_for_status()
+    return response.json()["cloud_accounts"]
```
antoniodimariano, line 32: Do you think it would be good to check if  `ctx.obj.opt_api_base_url` or  `ctx.obj.opt_cluster_secret` are None/empty before proceeding? I was thinking of changing this to 
```
base_url = ctx.obj.opt_api_base_url
cluster_secret = ctx.obj.opt_cluster_secret

if not base_url or not cluster_secret:
     raise ValueError("....")

Comment (antoniodimariano) [reply app/commands/redeem_entitlements.py:32]: Is the idea to raise an exception if the `cloud_accounts` doesn't exist?
`return response.json()["cloud_accounts"]`

File: app/commands/redeem_entitlements.py [hunk 2]
```diff
@@ -0,0 +1,114 @@
...
+            try:
+                datasources = fetch_datasources_for_organization(
+                    ctx,
+                    organization.operations_external_id,  # type: ignore
+                )
+            except httpx.HTTPError as e:
+                console.print(f"[red]Failed to fetch datasources: {e}[/red]")
+                continue
+            for datasource in datasources:
+                datasource_id = datasource["account_id"]
```
antoniodimariano, line 68: is the key `account_id` always defined?

app/commands/redeem_entitlements.py
ffaraone: When Settings load eventually raises an exception if no env has been set for opt_cluster_secret or opt_api_base_url so in this point they are not None for sure.
If the `cloud_accounts` can exist or not in the response it make sense to check but the response at most return a cloud_accounts = [] so there shouldn't be any problem here.

antoniodimariano (approved):

app/commands/redeem_entitlements.py
```diff
@@ -0,0 +1,114 @@
...
+                    Entitlement.status == EntitlementStatus.NEW,
+                )
+                instance = query.scalar()
+                if instance:
+                    instance.status = EntitlementStatus.ACTIVE
+                    instance.redeemed_at = datetime.now(UTC)
+                    instance.redeemed_by = organization
+                    instance.operations_external_id = datasource["id"]
+                    session.add(instance)
+                    session.commit()
```
arturbalabanov, line 105: Can we add exception handling here too, so that if there is an error the rest can continue?

Also, maybe we can move the commit at the end to update all of them at once? But that depends on whether we want to do that anyway

app/commands/redeem_entitlements.py
```diff
@@ -0,0 +1,114 @@
...
+                            "\t[orange3]Unsupported datasource type "
+                            f"{datasource["type"]}, skip it[/orange3]"
+                        )
+                        continue
+
+                query = session.query(Entitlement).where(
+                    Entitlement.datasource_id == datasource_id,
+                    Entitlement.status == EntitlementStatus.NEW,
+                )
+                instance = query.scalar()
```
arturbalabanov, line 98: No need to change anything as this works and this is just a script anyway but FYI:

* `session.query` is the old sqlalchemy 1.x interface (kept for backwards compatibility) and is deprecated in favour of `select`
* I recently learned that you can directly do `session.scalars(select(...))` as a shorthand of execute + scalars. There is also `session.scalar` if you're expecting only one result

app/commands/redeem_entitlements.py
```diff
@@ -0,0 +1,114 @@
...
+                if instance:
+                    instance.status = EntitlementStatus.ACTIVE
+                    instance.redeemed_at = datetime.now(UTC)
+                    instance.redeemed_by = organization
+                    instance.operations_external_id = datasource["id"]
```
arturbalabanov, line 103: We should set `updated_at` (and `updated_by`?) here too, right? Maybe we can use the entitlement handler instead, so that these fields are applied automatically

app/commands/redeem_entitlements.py
ffaraone: in theory yes!

arturbalabanov (approved):

app/commands/redeem_entitlements.py
ffaraone: About try/except -> sure MPT-7009
About commit at the end -> No it is a long running process and there is no requirement to have all or none redeemed.

app/commands/redeem_entitlements.py
ffaraone: Will change it in scope of MPT-7009

app/commands/redeem_entitlements.py
ffaraone: updated_at -> I don't know (read below)
updated_by -> we do not have an Actor that represent the this API itself.

PR #65, opened 2025-02-14
MPT-6999 output invited user id, output enum value for account type c…

ffaraone:
…apitalized

### DIFF
--- a/app/commands/invite_user.py
+++ b/app/commands/invite_user.py
@@ -116,9 +116,9 @@ def command(
         session.refresh(account_user)
 
         print(f"""
-[{color}]User [bold]{user.name} ({user.email})[/bold] {action}[/{color}]
+[{color}]User [bold]{user.id} - {user.name} ({user.email})[/bold] {action}[/{color}]
 
-Account: [blue][bold]{account.id}[/bold] - {account.name} ({account.type})[/blue]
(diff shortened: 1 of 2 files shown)


app/commands/invite_user.py
```diff
@@ -116,9 +116,9 @@ def command(
         session.refresh(account_user)
 
         print(f"""
-[{color}]User [bold]{user.name} ({user.email})[/bold] {action}[/{color}]
+[{color}]User [bold]{user.id} - {user.name} ({user.email})[/bold] {action}[/{color}]
 
-Account: [blue][bold]{account.id}[/bold] - {account.name} ({account.type})[/blue]
+Account: [blue][bold]{account.id}[/bold] - {account.name} ({account.type.value.title()})[/blue]
```
arturbalabanov, line 121: a handy trick with rich: you can just use `[/]` as a closing marker, it's especially useful when you have dynamic colours like here. And you can combine multiple markups too, here's an example:

```
this is normal text [bold blue]this is in bold blue[/] this is normal text again
```

obviously no need for a change here, just fyi :)

arturbalabanov (approved):

app/commands/invite_user.py
ffaraone: 👍

PR #67, opened 2025-02-17
Mpt 6547 account list users endpoint

antoniodimariano:
Add GET /accounts/account.id/users

### DIFF
--- a/app/routers/accounts.py
+++ b/app/routers/accounts.py
@@ -2,11 +2,19 @@ from typing import Annotated
 
 from fastapi import APIRouter, Depends, HTTPException, status
 from fastapi_pagination.limit_offset import LimitOffsetPage
+from sqlalchemy import select, exists
+from sqlalchemy.orm import joinedload
 
 from app.auth.auth import check_operations_account
 from app.db.handlers import NotFoundError
-from app.db.models import Account
-from app.dependencies import AccountId, AccountRepository, CurrentAuthContext, UserId
+from app.db.models import Account, User, AccountUser
+from app.dependencies import (
+    AccountId,
+    AccountRepository,
+    CurrentAuthContext,
+    UserId,
+    UserRepository,
+)
 from app.enums import AccountStatus, AccountType
 from app.pagination import paginate
 from app.schemas import (
@@ -14,6 +22,7 @@ from app.schemas import (
     AccountRead,
     AccountUpdate,
     AccountUserRead,
+    UserRead,
     from_orm,
     to_orm,
 )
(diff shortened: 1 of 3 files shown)


app/routers/accounts.py
```diff
@@ -189,15 +196,37 @@ async def update_account(
...
+            detail="Cheating is bad. Don't do it.",
+        )
+    return await paginate(
+        user_repo,
+        UserRead,
+        extra_conditions=[
+            exists().where(
+                AccountUser.account_id == account.id,
+                User.id == AccountUser.user_id,
+                User.status == UserStatus.ACTIVE,
```
arturbalabanov, line 226: Looking at the screenshot in the jira issue, I don't think we want to filter out non active users. @ffaraone can you please confirm? I'm not sure if that applies to all account types or only for operations

app/routers/accounts.py
```diff
@@ -189,15 +196,37 @@ async def update_account(
...
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="Cheating is bad. Don't do it.",
+        )
+    return await paginate(
+        user_repo,
+        UserRead,
+        extra_conditions=[
+            exists().where(
+                AccountUser.account_id == account.id,
+                User.id == AccountUser.user_id,
```
arturbalabanov, line 225: This line threw me off initially and thought it was a bug but I _think_ this is used for the JOIN condition, is that right? Just wanted to confirm, maybe adding a small comment would help future maintainers

tests/test_accounts_api.py
```diff
@@ -477,3 +466,134 @@ async def test_can_only_updated_the_name_and_external_id(
...
+    mock_account_repo.get.assert_awaited_once_with(id="invalid_account_id")
+
+
+async def test_get_all_list_account_users_multiple_pages(
+    operations_client: AsyncClient,
+    operations_account: Account,
+    accountuser_factory: ModelFactory[AccountUser],
+    user_factory: ModelFactory[User],
+    db_session: AsyncSession,
+):
```
arturbalabanov, line 546: Not a big deal but this test would really benefit from `pytest.mark.parametrize` -- not only will it be shorter but also if one of the page assertions fails we'll still see the results for the other pages and it will be easier to track down the  bug. I've been making use of parametrization a lot in the systems api tests, you can check them out for examples

No need for a change here as the test is doing its job as it is, just sharing knowledge :)

app/routers/accounts.py
antoniodimariano: yes it's a JOIN condition like 
stmt = (
        select(User)
        .join(AccountUser, User.id == AccountUser.user_id)
        .where(AccountUser.account_id == account.id)
    )

app/routers/accounts.py
ffaraone: Yeah you are right @arturbalabanov, they shouldn't be filtered out.

ffaraone (changes requested):

tests/test_accounts_api.py
antoniodimariano: I agree. I also use it when it comes to test different values. Probably it can be done here too.

app/routers/accounts.py
antoniodimariano: In the task's description I see 
```
Description

GET /accounts/{id}/users
The API must provide an endpoint to list all the Accounts the current logged user belongs to.
 
 
The endpoint can only be consumed if the user is active. [...]
```

arturbalabanov (approved):

app/routers/accounts.py
arturbalabanov: That's so much more clear, thanks! This is on me tbh, the handers look less and less useful as we use them, I foresee a big refactoring happening in the future :D

app/routers/accounts.py
```diff
@@ -189,15 +196,40 @@ async def update_account(
...
+        - HTTPException with status 400 if the given account is different from the context account
+        - HTTPException 404 if the provided account's id doesn't exist.
+    Returns a list of accounts if any.
     """
-    pass
+
+    if auth_context.account.type == AccountType.AFFILIATE and auth_context.account != account:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="Cheating is bad. Don't do it.",
```
ffaraone, line 217: For a 403 we can use a funny message but for a 400 we should use something meaningful. Sorry to not see it before.

ffaraone (changes requested):

PR #73, opened 2025-02-18
feat: Create system endpoint (MPT-6527)

arturbalabanov:


app/routers/systems.py
```diff
@@ -64,8 +64,44 @@ async def get_systems(
...
 @router.post("", response_model=SystemRead, status_code=status.HTTP_201_CREATED)
-async def create_system(data: SystemCreate):  # pragma: no cover
-    pass
+async def create_system(
+    data: SystemCreate,
+    account_repo: AccountRepository,
+    system_repo: SystemRepository,
+    auth_ctx: CurrentAuthContext,
+):
+    if auth_ctx.account.type == AccountType.AFFILIATE and data.owner.id != auth_ctx.account.id:
```
ffaraone, line 73: Like we have for Entitlements, data.owner is mandatory only for payloads sent by an Operations account and must not be included in payloads sent by an Affiliate account. 
Could you please implement it this way ?

app/routers/systems.py
```diff
@@ -64,8 +64,44 @@ async def get_systems(
...
+                status=SystemStatus.ACTIVE,
+            )
+        )
+    except ConstraintViolationError as e:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="A system with the same external ID already exists.",
+        ) from e
+
+    return from_orm(SystemRead, system)
```
antoniodimariano, line 104: I'd use just one `try - except` block for better readability  and also use some logger messages to facilitate the debugging 
```
 try:
        owner_account = await account_repo.get(data.owner.id)
        
        system = await system_repo.create(
            System(
                name=data.name,
                description=data.description,
                external_id=data.external_id,
                jwt_secret=data.jwt_secret,
                owner=owner_account,
                status=SystemStatus.ACTIVE,
            )
        )

    except NotFoundError as e:
        logger.warning(f"The owner account does not exist: {data.owner.id}")
        raise HTTPException(
            detail="The owner account does not exist.",
        ) from e

    except ConstraintViolationError as e:
        logger.warning(f"A system with the same external ID already exists.: {data.external_id}")
        raise HTTPException(
            detail="A system with the same external ID already exists.",
        ) from e

    return from_orm(SystemRead, system)
    ```

app/routers/systems.py
ffaraone: The JWT token should be returned only as the response of the create endpoint and of the update in the case it is received in the request. In all the other endpoints it should be omitted. Could you please check that ?

PR #75, opened 2025-02-20
feat: Improve system endpoints payloads (MPT-7106)

arturbalabanov:
* system create: `data.owner` is mandatory only for Operators and is forbidden for affiliates (just like entitlements)
* JWT token is returned only in the system create response (incl. tests that we don't accidentally return it in other endpoints responses)
* Add logging and improve readability when handling exceptions in routes

### DIFF
--- a/app/routers/systems.py
+++ b/app/routers/systems.py
@@ -9,7 +9,8 @@ from app.db.models import System
 from app.dependencies import AccountRepository, CurrentAuthContext, SystemId, SystemRepository
 from app.enums import AccountType, SystemStatus
 from app.pagination import paginate
-from app.schemas import SystemCreate, SystemRead, SystemUpdate, from_orm
+from app.schemas import SystemCreate, SystemCreateResponse, SystemRead, SystemUpdate, from_orm
+from app.utils import wrap_exc_in_http_response
 
 # ============
 # Dependencies
@@ -33,13 +34,8 @@ async def fetch_system_or_404(
     system_repo: SystemRepository,
     extra_conditions: CommonConditions,
 ) -> System:
-    try:
+    async with wrap_exc_in_http_response(NotFoundError, status_code=status.HTTP_404_NOT_FOUND):
         return await system_repo.get(id=id, extra_conditions=extra_conditions)
-    except NotFoundError as e:
-        raise HTTPException(
-            status_code=status.HTTP_404_NOT_FOUND,
-            detail=str(e),
-        ) from e
 
 
 # ======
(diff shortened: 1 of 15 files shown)


app/routers/systems.py
```diff
@@ -33,13 +34,8 @@ async def fetch_system_or_404(
     system_repo: SystemRepository,
     extra_conditions: CommonConditions,
 ) -> System:
-    try:
+    async with wrap_exc_in_http_response(NotFoundError, status_code=status.HTTP_404_NOT_FOUND):
```
ffaraone, line 37: this context managers are not async, I could understand the preference of a context manager vs a try/except block for readability even if we are not managing any context but we should at least make them sync, wdyt ?

antoniodimariano (approved):

app/routers/systems.py
```diff
@@ -63,28 +59,28 @@ async def get_systems(
...
     account_repo: AccountRepository,
     system_repo: SystemRepository,
     auth_ctx: CurrentAuthContext,
 ):
-    if auth_ctx.account.type == AccountType.AFFILIATE and data.owner.id != auth_ctx.account.id:
-        raise HTTPException(
-            status_code=status.HTTP_400_BAD_REQUEST,
-            detail="Affiliate users can only create systems bound to their own account.",
-        )
+    if data.owner is None:
```
ffaraone, line 69: I think operations account must always send an explicit owner

ffaraone (changes requested):

app/utils.py
```diff
@@ -13,3 +16,22 @@ async def wrap_http_error_in_502(base_msg: str = "Error in FinOps for Cloud"):
...
+
+
+@contextlib.asynccontextmanager
+async def wrap_exc_in_http_response(
+    exc_cls: type[Exception],
+    error_msg: str | None = None,
+    status_code: int = status.HTTP_400_BAD_REQUEST,
+):
+    try:
+        yield
```
ffaraone, line 28: this context managers should be sync since we do not perform asyncio neither in `__aenter__` nor in `__aexit__`

ffaraone (changes requested):

app/routers/systems.py
arturbalabanov: Good spot, you're right, converted it to regular context manager and converted other unnecessarily async functions to sync

pyproject.toml
```diff
@@ -70,16 +70,19 @@ extend-exclude = [
 cache-dir = ".cache/ruff"
 
 [tool.ruff.lint]
+preview = true # enable linting rules in preview (e.g. RUF029 as of 2025-02-20)
+explicit-preview-rules = true # only enable preview rules we explicitly specify
```
arturbalabanov, line 74: Leaving this discussion open until we all agree this is a good idea, otherwise I'll remove it, no worries

taplo.toml
arturbalabanov, line 1: Started using this tool to lint and format toml files. Definitely not pushing anyone else to use it but had to add this config option, otherwise it was formatting this:

```toml
exclude_lines = [
    "pragma: no cover",
    "pragma: no branch",
    "NotImplementedError",
]
```

to:

```toml
exclude_lines = ["pragma: no cover", "pragma: no branch", "NotImplementedError"]
```

which is less readable imo

ffaraone (approved):

PR #81, opened 2025-02-21
MPT-7277 support limit=0 for getting the count

ffaraone:


veralimita (approved):

arturbalabanov (approved):
Approved as it's working as intended.

But it looks a bit weird to for an endpoint to support that, it's not intuitive that it should be used to get the count. Maybe we can have a separate endpoint `/count` endpoint for that? I don't feel too strongly about this though, just a thought

PR #83, opened 2025-02-21
chore: group audit events in the responses (only created, updated and deleted for now) (MPT-7140)

arturbalabanov:
I'm honestly not very happy with this implementation, looks hacky but it was the simplest way I could think of which didn't
require large refactoring of our models or using more obscure Pydantic models features. After spending more than a day on this had to draw the line somewhere and at least this is working.

Also left a few TODO comments for events other than created, updated and deleted. I didn't know if we want to include them too but if we do, that will likely require bigger changes as this hack may not work for them.

We can either merge it as it is and kick the bucket down the road when we have more time to do it properly,
or decide to leave the PR open until we have time to do it properly. I'm fine with both though I'm leaning towards the first
option as the merge conflicts later on will be a lot harder to resolve.


app/schemas.py
```diff
@@ -84,18 +113,23 @@ class ActorRead(ActorBase, IdSchema):
...
+
+
 class CommonEventsSchema(BaseSchema):
-    created_at: datetime.datetime
-    updated_at: datetime.datetime
-    deleted_at: datetime.datetime | None = None
-    created_by: ActorReference | None = None
-    updated_by: ActorReference | None = None
-    deleted_by: ActorReference | None = None
+    events: AuditEventsSchema
```
ffaraone, line 132: I think here it should be a dict of [str, AuditFieldSchema].

app/schemas.py
```diff
@@ -84,18 +113,23 @@ class ActorRead(ActorBase, IdSchema):
...
+class ActorReference(ActorBase, IdSchema):
     name: Annotated[str, Field(examples=["Barack Obama"])]
 
 
+class AuditFieldSchema(BaseSchema):
+    at: datetime.datetime
+    by: ActorReference | None
+
+
+class AuditEventsSchema(BaseSchema):
```
ffaraone, line 125: This doesn't allow us to support the `terminated` event of an Entitlement.

app/schemas.py
```diff
@@ -22,8 +29,30 @@
...
-    return cls.model_validate(db_model)
+def from_orm[M: Base, S: BaseModel](schema_cls: type[S], db_model: M) -> S:
+    if not issubclass(schema_cls, CommonEventsSchema):
+        return schema_cls.model_validate(db_model)
+
+    # NOTE: This is a hack, ideally this behaviour should be handled by the models,
+    #       but to do it properly we need to spend more time in learning how pydantic
+    #       works and possibly do quite a lot of refactoring of our schemas
+
+    events = AuditEventsSchema(
```
ffaraone, line 40: I don't think we can generalize here or at least not that way. For example, Entitlement object support an extra event "terminated".

ffaraone (changes requested):

app/schemas.py
arturbalabanov: You're right, however if we don't do that here (or somewhere in the models but again -- that's generic), then we need to change almost all endpoints to specifically set that data as they all use `from_orm` and that's not only a lot of changes to review in one go but also makes writing endpoints unnecessarily verbose. Will try to find a way to generalise this more to include more fields or maybe pass additional ones as arguments

ffaraone:
I would suggest a simplified implementation without using Generics:

```python
from __future__ import annotations

import datetime
import types
from typing import Annotated

from pydantic import (
    BaseModel,
    ConfigDict,
    Field,
    SecretStr,
    field_validator,
)

from app.db.models import Base
from app.enums import (
    ActorType,
)

def from_orm[M: Base, S: BaseModel](schema_cls: type[S], db_model: M) -> S:

    if not issubclass(schema_cls, CommonEventsSchema):
        return schema_cls.model_validate(db_model)

    events_schema_cls = schema_cls.model_fields["events"].annotation
    events = {}
    for field_name, field_info in events_schema_cls.model_fields.items():
        event_data_cls = field_info.annotation

        if isinstance(event_data_cls, types.UnionType):
            event_data_cls = event_data_cls.__args__[0]

        at_field = f"{field_name}_at"
        by_field = f"{field_name}_by"
        at_field_value = getattr(db_model, at_field)
        by_field_value = getattr(db_model, by_field)
        event = event_data_cls(
            at=at_field_value,
            by=by_field_value,
        )
        events[field_name] = event
    events_schema = events_schema_cls(**events)

    db_model.events = events_schema

    return schema_cls.model_validate(db_model)


def to_orm[M: Base, S: BaseModel](schema: S, model_cls: type[M]) -> M:
    """
    Converts a Pydantic schema instance to an ORM model instance.

    This function ensures that only fields present in the ORM model are
    passed to it,preventing errors caused by extra fields.
    """
    # extract data from the schema. This will return a dict
    schema_data = schema.model_dump(exclude_unset=True)
    # filter out all the fields that are not in the ORM model
    dbmodel_fields = {key: value for key, value in schema_data.items() if hasattr(model_cls, key)}
    # create an instance of the ORM model
    return model_cls(**dbmodel_fields)


class BaseSchema(BaseModel):
    model_config = ConfigDict(from_attributes=True, extra="ignore")


class IdSchema(BaseSchema):
    id: str


class ActorBase(BaseSchema):
    type: ActorType


class ActorRead(IdSchema, ActorBase):
    pass


class ActorReference(IdSchema):
    type: ActorType
    name: Annotated[str, Field(examples=["Barack Obama"])]


class EventSchema(BaseSchema):
    at: datetime.datetime | None
    by: ActorReference | None


class AuditEventsSchema(BaseSchema):
    created: EventSchema
    updated: EventSchema
    deleted: EventSchema | None = None


class CommonEventsSchema(BaseSchema):
    events: AuditEventsSchema

```

and for Entitlement:

```python
class EntitlementReedemEventSchema(BaseSchema):
    at: datetime.datetime
    by: OrganizationReference



class EntitlementsEventsSchema(AuditEventsSchema):
    redeemed: EntitlementReedemEventSchema | None = None
    terminated: EventSchema | None = None


class EntitlementRead(IdSchema, EntitlementBase):
    linked_datasource_id: Annotated[
        str | None, Field(max_length=255, examples=["ee7ebfaf-a222-4209-aecc-67861694a488"])
    ] = None
    linked_datasource_name: Annotated[
        str | None, Field(max_length=255, examples=["Azure Dev Subscription"])
    ] = None
    linked_datasource_type: DatasourceType | None = None
    events: EntitlementsEventsSchema
    owner: AccountReference
    status: EntitlementStatus
```

ffaraone (approved):

PR #87, opened 2025-02-24
Mpt 6623 user get details

antoniodimariano:
Add GET /users/{id}
Adjust commands test 
Add Test for /users/{id}

### DIFF
--- a/app/routers/users.py
+++ b/app/routers/users.py
@@ -259,11 +259,47 @@ async def resend_user_invitation(
 
 
 @router.get("/{id}", response_model=UserRead)
-async def get_user_by_id(id: str, token: str | None = None):  # pragma: no cover
+async def get_user_by_id(
+    id: str,
+    auth_context: CurrentAuthContext,
+    accountuser_repo: AccountUserRepository,
+    user_repo: UserRepository,
+    token: str | None = None,
+):  # pragma: no cover
     # if token is provided no authentication is needed but
     # an AccountOperator in status invited must exist with
     # user id and token and the token must not be expired
-    pass  # not yet implemented
+
+    """
+    If Access_Token
+    1. operations => ritorno utente
+    2. affiliate =>  authcontext AccountUser e vedere se esiste un record per
+    {userid}, l'account preso dell'autocontext e status != deleted .
+
+
+    If No Access Token but there is an Invitation Token
+
+    controllare in BD AccountUser per verificare invitation token e {id}
+    Se nn esiste -> 401
+
+
+    """
+    user_id = id
+    if auth_context.account.type == AccountType.OPERATIONS:
+        return await user_repo.get(id=user_id)
+    elif auth_context.account.type == AccountType.AFFILIATE:
+        account_user = await accountuser_repo.get_account_user(
+            account_id=auth_context.account.id,
+            user_id=user_id,
+            extra_conditions=[AccountUser.status != AccountUserStatus.DELETED],
+        )
+        if account_user is None:
+            logger.error("No account user has been found")
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail="Account Not Found",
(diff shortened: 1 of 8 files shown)


app/routers/users.py
```diff
@@ -259,11 +262,66 @@ async def resend_user_invitation(
...
+            AccountUser.invitation_token == invitation_token,
+            AccountUser.status.in_(
+                [AccountUserStatus.INVITED, AccountUserStatus.INVITATION_EXPIRED]
+            ),
+        )
+        if account_user is None:
+            raise UNAUTHORIZED_EXCEPTION
+        response = await user_repo.get(id=user_id)
+    elif auth_context.account.type == AccountType.OPERATIONS:
+        with wrap_exc_in_http_response(NotFoundError, "User Not Found."):
```
ffaraone, line 304: In this case NotFoundError bring a standard message used then by `wrap_exc_in_http_response` (see [https://github.com/softwareone-platform/ffc-finops-operations/blob/main/app/utils.py#L22C5-L22C30](https://github.com/softwareone-platform/ffc-finops-operations/blob/main/app/utils.py#L22). What you have to pass to the `wrap_exc_in_http_response` is the status code you want to use since by default it use a 400. See [https://github.com/softwareone-platform/ffc-finops-operations/blob/main/app/routers/systems.py#L37](https://github.com/softwareone-platform/ffc-finops-operations/blob/main/app/routers/systems.py#L37)

app/routers/users.py
```diff
@@ -259,11 +262,66 @@ async def resend_user_invitation(
...
+            extra_conditions=[AccountUser.status != AccountUserStatus.DELETED],
+        )
+        if account_user is None:
+            logger.error("No account user has been found.")
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail="User Not Found.",
+            )
+        with wrap_exc_in_http_response(NotFoundError, "User Not Found."):
+            # not filtering [AccountUser.status != AccountUserStatus.DELETED]
```
ffaraone, line 319: Same as above here!

app/routers/users.py
```diff
@@ -259,11 +262,66 @@ async def resend_user_invitation(
...
+        account_user = await accountuser_repo.get_account_user(
+            account_id=auth_context.account.id,
+            user_id=user_id,
+            extra_conditions=[AccountUser.status != AccountUserStatus.DELETED],
+        )
+        if account_user is None:
+            logger.error("No account user has been found.")
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail="User Not Found.",
```
ffaraone, line 316:
```suggestion
                detail=f"User with ID `{user_id}` wasn't found."),
```

ffaraone (changes requested):

app/routers/users.py
```diff
@@ -259,11 +262,70 @@ async def resend_user_invitation(
...
+            extra_conditions=[AccountUser.status != AccountUserStatus.DELETED],
+        )
+        if account_user is None:
+            logger.error("No account user has been found.")
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail=f"User with ID `{user_id}` wasn't found.",
+            )
+        with wrap_exc_in_http_response(
+            NotFoundError, status_code=status.HTTP_404_NOT_FOUND, error_msg="User Not Found."
```
ffaraone, line 321: Sorry I wasn't able to explain: you shouldn't pass the error_msg since the NotFoundError already contains the right message, so if you don't pass it the error message will be the same as above (f"User with ID `{user_id}` wasn't found.").


```suggestion
            NotFoundError, status_code=status.HTTP_404_NOT_FOUND
```

ffaraone (changes requested):

ffaraone (approved):

PR #90, opened 2025-02-26
feat: Allow partial updates when using systems PUT endpoint (MPT-7365)

arturbalabanov:


ffaraone:
@arturbalabanov do we have to check for the case where no field have been passed for the update ?

arturbalabanov:
> @arturbalabanov do we have to check for the case where no field have been passed for the update ?

Good point, added a check and a test for that

PR #93, opened 2025-02-26
Add /users Endpoints and related logic.

antoniodimariano:
Add /users Endpoints and related logic. 
Refactor of from_orm and to_orm

### DIFF
--- a/app/db/models.py
+++ b/app/db/models.py
@@ -190,6 +190,7 @@ class User(Actor, HumanReadablePKMixin, AuditableMixin):
         try:
             return self.accounts[0]
         except Exception:
+            # todo : improve this excepption
             return None
 
     __mapper_args__ = {
--- a/app/db/models.py
+++ b/app/db/models.py
@@ -185,6 +185,13 @@ class User(Actor, HumanReadablePKMixin, AuditableMixin):
     )
     accounts: Mapped[list["AccountUser"]] = relationship(back_populates="user")
 
+    @property
+    def account_user(self):
+        try:
+            return self.accounts[0]
+        except Exception:
+            return None
+
     __mapper_args__ = {
         "polymorphic_identity": ActorType.USER.value,
         "inherit_condition": id == Actor.id,
--- a/app/pagination.py
+++ b/app/pagination.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from collections.abc import Sequence
 from typing import TypeVar
 
@@ -8,10 +10,11 @@ from fastapi_pagination.limit_offset import LimitOffsetPage as _LimitOffsetPage
 from fastapi_pagination.types import GreaterEqualZero
 from pydantic import BaseModel
 from sqlalchemy import ColumnExpressionArgument
+from sqlalchemy.orm.interfaces import ORMOption
 
 from app.db.handlers import ModelHandler
 from app.db.models import Base
-from app.schemas.core import BaseSchema, from_orm
+from app.schemas.core import BaseSchema, convert_model_to_schema
 
 T = TypeVar("T")
 
@@ -38,19 +41,26 @@ async def paginate[M: Base, S: BaseSchema](
     schema_cls: type[S],
     *,
     extra_conditions: list[ColumnExpressionArgument] | None = None,
+    options: list[ORMOption] | None = None,
 ) -> AbstractPage[S]:
+    """
+    This function queries a database model (M) using a ModelHandler.
+    It applies optional filtering (extra_conditions) and query options (options).
+    It then serializes the results into a schema (S) and returns
+    a paginated response in the form of AbstractPage[S].
+    """
     params: LimitOffsetParams = resolve_params()
-    extra_conditions = extra_conditions or []
-
-    total = await handler.count(*extra_conditions)
+    total = await handler.count(*(extra_conditions or []))
     items: Sequence[M] = []
     if params.limit > 0:
         items = await handler.fetch_page(
-            limit=params.limit, offset=params.offset, extra_conditions=extra_conditions
+            limit=params.limit,
+            offset=params.offset,
+            extra_conditions=extra_conditions,
+            options=options,
         )
-
     return create_page(
-        [from_orm(schema_cls, item) for item in items],
+        [convert_model_to_schema(schema_cls, item) for item in items],
         params=params,
         total=total,
     )
--- a/app/pagination.py
+++ b/app/pagination.py
@@ -41,7 +41,7 @@ async def paginate[M: Base, S: BaseSchema](
     schema_cls: type[S],
     *,
     extra_conditions: list[ColumnExpressionArgument] | None = None,
-    options: list[ORMOption] | None = None,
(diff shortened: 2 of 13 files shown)


app/schemas/core.py
```diff
@@ -11,63 +11,110 @@
...
-            continue
+        schema_values[field_name] = (
+            event_field_schema_cls(at=at_value, by=by_value) if at_value else None
+        )  # noqa: E501
+    return events_schema_cls(**schema_values)
 
-        by_value = getattr(db_model, f"{field_name}_by", None)
 
-        schema_values[field_name] = event_field_schema_cls(at=at_value, by=by_value)
+def resolve_field_type(field_info: Any) -> Any:
```
ffaraone, line 92:
```suggestion
def resolve_field_type(field_info: Any) -> BaseModel:
```

app/schemas/core.py
```diff
@@ -11,63 +11,110 @@
...
 
+        if len(non_none_types) != 1:
+            raise TypeError(f"Unsupported union type: {event_field_schema_cls.__args__}")
 
-def to_orm[M: Base, S: BaseModel](schema: S, model_cls: type[M]) -> M:
+        event_field_schema_cls = non_none_types[0]
+    return event_field_schema_cls
+
+
+def convert_schema_to_model(schema: S, model_cls: type[M]) -> M:
```
ffaraone, line 117:
```suggestion
def convert_schema_to_model[M: Base, S: BaseModel](schema: S, model_cls: type[M]) -> M:
```

app/schemas/core.py
```diff
@@ -82,6 +129,16 @@ def to_orm[M: Base, S: BaseModel](schema: S, model_cls: type[M]) -> M:
     return model_cls(**dbmodel_fields)
 
 
+def extract_fields_from_model(
```
ffaraone, line 132:
```suggestion
def extract_fields_from_model[M: Base, S: BaseModel](
```

app/schemas/core.py
```diff
@@ -82,6 +129,16 @@ def to_orm[M: Base, S: BaseModel](schema: S, model_cls: type[M]) -> M:
...
+    schema_cls: type[S], db_model: M, excluded_fields: list
```
ffaraone, line 133:
```suggestion
    schema_cls: type[S], db_model: M, excluded_fields: list[str]
```

ffaraone (changes requested):

app/db/models.py
```diff
@@ -185,6 +185,13 @@ class User(Actor, HumanReadablePKMixin, AuditableMixin):
     )
     accounts: Mapped[list["AccountUser"]] = relationship(back_populates="user")
 
+    @property
+    def account_user(self):
```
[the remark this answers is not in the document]
antoniodimariano: Agree with you about the importance of naming.

app/db/models.py
```diff
@@ -185,6 +185,13 @@ class User(Actor, HumanReadablePKMixin, AuditableMixin):
...
+        try:
+            return self.accounts[0]
+        except Exception:
+            return None
```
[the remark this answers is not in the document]
antoniodimariano: Yes, that `Exception` is too broad, and we were well aware of that when we discussed it with Francesco. 
It has to be changed to something more meanful

app/db/models.py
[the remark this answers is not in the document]
ffaraone: Agree also, but in a separate PR.

app/db/models.py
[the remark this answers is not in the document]
ffaraone: AFAIK it is not possible. As soon as you access `.accounts` try to fetch them from the database. If the object is not bound to a session it fails (with a very ugly exception related to greenlets).
For sure we should use a better exception here but I'm still trying to find which one should be used.

Once we find it, we can also switch to `contextlib.suppress` since we already used it in handlers and IMO is a bit more pythonic ;)

app/pagination.py
```diff
@@ -8,10 +10,11 @@
...
 from pydantic import BaseModel
 from sqlalchemy import ColumnExpressionArgument
+from sqlalchemy.orm.interfaces import ORMOption
 
 from app.db.handlers import ModelHandler
 from app.db.models import Base
-from app.schemas.core import BaseSchema, from_orm
+from app.schemas.core import BaseSchema, convert_model_to_schema
 
 T = TypeVar("T")
```
ffaraone, line 19: What is this ?

app/pagination.py
```diff
@@ -38,19 +41,26 @@ async def paginate[M: Base, S: BaseSchema](
...
+    total = await handler.count(*(extra_conditions or []))
     items: Sequence[M] = []
     if params.limit > 0:
         items = await handler.fetch_page(
-            limit=params.limit, offset=params.offset, extra_conditions=extra_conditions
+            limit=params.limit,
+            offset=params.offset,
+            extra_conditions=extra_conditions,
+            options=options,
         )
```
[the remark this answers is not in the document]
antoniodimariano: Are you thinking of passing a dict?

app/routers/users.py
```diff
@@ -256,8 +291,11 @@ async def resend_user_invitation(
         days=settings.invitation_token_expires_days
     )
     account_user = await accountuser_repository.update(account_user)
-    response = from_orm(UserInvitationRead, user)
-    response.account_user = from_orm(AccountUserRead, account_user)
+    response = convert_model_to_schema(
+        UserInvitationRead,
+        user,
+        account_user=convert_model_to_schema(AccountUserRead, account_user),
+    )  # noqa: E501
```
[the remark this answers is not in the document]
ffaraone: 👍

app/pagination.py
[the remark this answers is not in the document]
arturbalabanov: You mean like `**kwargs`? You could do that, my comment was just to rename `options` to `page_options` to make it clear it's only used in the get page query and not the count one (as opposed to `extra_conditions` which is passed to both)

ffaraone (changes requested):

app/pagination.py
```diff
@@ -38,19 +41,26 @@ async def paginate[M: Base, S: BaseSchema](
...
-    total = await handler.count(*extra_conditions)
+    total = await handler.count(*(extra_conditions or []))
     items: Sequence[M] = []
     if params.limit > 0:
         items = await handler.fetch_page(
-            limit=params.limit, offset=params.offset, extra_conditions=extra_conditions
+            limit=params.limit,
+            offset=params.offset,
+            extra_conditions=extra_conditions,
+            page_options=options,
```
ffaraone, line 60: in paginate the name of the function argument should be page_options, in the handler either options either orm_options. Is it what you requested @arturbalabanov  ?

app/pagination.py
antoniodimariano: do you mean the `convert_model_to_schema` function ?

ffaraone: My mistake, definition of T is there from before. I will look to change it.

ffaraone (approved):

PR #94, opened 2025-02-27
Add POST /users/{id}/disable

antoniodimariano:
Add POST /users/{id}/disable

### DIFF
--- a/app/routers/users.py
+++ b/app/routers/users.py
@@ -223,9 +223,9 @@ async def get_user_accounts(id: str):  # pragma: no cover
 
 @router.post(
     "/{id}/disable",
-    dependencies=[Depends(authentication_required)],
+    dependencies=[Depends(check_operations_account)],
     response_model=UserRead,
-    status_code=status.HTTP_201_CREATED,
+    status_code=status.HTTP_200_OK,
 )
 async def disable_user(
     id: str,
@@ -237,11 +237,7 @@ async def disable_user(
     This endpoint allows an OPERATOR to disable a user.
     A user cannot disable itself.
     """
-    if auth_context.account.type != AccountType.OPERATIONS:
-        raise HTTPException(
-            status_code=status.HTTP_401_UNAUTHORIZED,
-            detail="You are not authorized to perform this action.",
-        )
+
     if user == auth_context.user:
         raise HTTPException(
             status_code=status.HTTP_400_BAD_REQUEST,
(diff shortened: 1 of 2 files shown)


app/routers/users.py
```diff
@@ -221,9 +225,37 @@ async def get_user_accounts(id: str):  # pragma: no cover
     "/{id}/disable",
     dependencies=[Depends(authentication_required)],
     response_model=UserRead,
+    status_code=status.HTTP_201_CREATED,
```
ffaraone, line 228: This must be a 200 not 201.

app/routers/users.py
```diff
@@ -221,9 +225,37 @@ async def get_user_accounts(id: str):  # pragma: no cover
...
+    id: str,
+    user_repo: UserRepository,
+    auth_context: CurrentAuthContext,
+    user: Annotated[User, Depends(fetch_user_or_404)],
+):
+    """
+    This endpoint allows an OPERATOR to disable a user.
+    A user cannot disable itself.
+    """
+    if auth_context.account.type != AccountType.OPERATIONS:
```
ffaraone, line 240: We already have this [https://github.com/softwareone-platform/ffc-finops-operations/blob/main/app/auth/auth.py#L143](https://github.com/softwareone-platform/ffc-finops-operations/blob/main/app/auth/auth.py#L143).

app/routers/users.py
```diff
@@ -221,9 +225,37 @@ async def get_user_accounts(id: str):  # pragma: no cover
...
+    if user == auth_context.user:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="A user cannot disable itself.",
+        )
+    if user.status != UserStatus.ACTIVE:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=(
+                f"User's status is '{user.status._value_}'; " "only active users can be disabled."
```
ffaraone, line 254: Please join this two string in one (😇 pycharm). Please use reversed single quotes (like we have in other places).

ffaraone (changes requested):

app/routers/users.py
antoniodimariano: yep ...

app/routers/users.py
antoniodimariano: it's a POST, that's 201

ffaraone: It is 201 only for create endpoint. This endpoint doesn't create an object.

PR #97, opened 2025-02-28
feat: Add endpoint to list accounts linked to a given user (MPT-6624)

arturbalabanov:


app/routers/users.py
```diff
@@ -215,10 +228,27 @@ async def delete_user(id: str):  # pragma: no cover
...
+    user: Annotated[User, Depends(fetch_user_or_404)],
+    accountuser_repo: AccountUserRepository,
+    auth_ctx: CurrentAuthContext,
+):
+    extra_conditions: list[ColumnExpressionArgument] = [
+        AccountUser.user == user,
+    ]
+    if auth_ctx is not None and auth_ctx.account.type == AccountType.AFFILIATE:
+        extra_conditions += [
+            AccountUser.account.has(Account.status != AccountStatus.DELETED),
```
ffaraone, line 243: This condition is not needed.
For the moment there is no possibility to delete an account.
In the future, I expect that, deleting an account cascade on its AccountUser

app/routers/users.py
```diff
@@ -215,10 +228,27 @@ async def delete_user(id: str):  # pragma: no cover
 @router.get(
     "/{id}/accounts",
     dependencies=[Depends(authentication_required)],
-    response_model=list[AccountUserRead],
+    response_model=LimitOffsetPage[AccountUserRead],
```
ffaraone, line 231: I think this endpoint should return an Account not AccountUser.

ffaraone (changes requested):

app/routers/users.py
ffaraone: Forget to mention that of course we need information about the account user but it should be nested

app/routers/users.py
ffaraone: In any case if we return a page of Accounts, the query will be done on the Account table joining the AccountUser

tests/api/users/test_collection.py
```diff
@@ -962,3 +964,177 @@ async def test_reset_password_waek_password(
...
     assert detail["msg"] == f"Value error, {expected_msg}."
+
+
+# ---------------------
+# GET ACCOUNTS FOR USER
+# ---------------------
+
+
+@dataclass
+class GetAccountsForUserTestCase:
```
ffaraone, line 975: Using a dataclass to parametrize tests it's a bit brain damaging for me

app/routers/users.py
```diff
@@ -264,10 +272,27 @@ async def delete_user(
...
+    if auth_ctx is not None and auth_ctx.account.type == AccountType.AFFILIATE:
+        extra_conditions.append(
+            Account.users.any(AccountUser.status != AccountUserStatus.DELETED),
+        )
+
+    return await paginate(
+        account_repo,
+        AccountRead,
+        extra_conditions=extra_conditions,
+        page_options=[joinedload(Account.users)],
```
ffaraone, line 294: You have to also add the `with_loader_criteria`  otherwise this joinedload will load all the AccountUser related to this acount

ffaraone (changes requested):

app/schemas/accounts.py
```diff
@@ -35,12 +35,17 @@ class AccountUpdate(BaseSchema):
...
-    status: AccountStatus
+class AccountReference(IdSchema):
+    name: Annotated[str, Field(max_length=255, examples=["Microsoft"])]
     type: AccountType
 
 
-class AccountReference(IdSchema):
-    name: Annotated[str, Field(max_length=255, examples=["Microsoft"])]
+# Importing here to avoid circular imports
+from app.schemas.users import AccountUserReferenceWithUser  # noqa: E402
```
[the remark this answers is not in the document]
arturbalabanov: I tried to extract it but there are still circular dependancies as the account user depends on both account and user and they both use it somewhere.

I'm not sure even if we extract the references that this will fix it completely but that's an overkill IMO. Let's leave it as it is for now and we can address it in another PR. It doesn't bother me that much tbf 🤷

app/routers/users.py
arturbalabanov: Good point, fixed and added a test for it too

ffaraone (approved):

PR #99, opened 2025-02-28
Mpt 6523 user enable endpoint

antoniodimariano:
Mpt 6523 user enable endpoint

### DIFF
--- a/app/routers/users.py
+++ b/app/routers/users.py
@@ -220,13 +220,16 @@ async def update_user(
     status_code=status.HTTP_204_NO_CONTENT,
 )
 async def delete_user(
-    user_repo: UserRepository,
     accountuser_repo: AccountUserRepository,
     auth_context: CurrentAuthContext,
     db_engine: DBEngine,
     db_session: DBSession,
     user: Annotated[User, Depends(fetch_user_or_404)],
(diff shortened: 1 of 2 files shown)


ffaraone (changes requested):
In this PR there are the delete_user and the enable_user endpoints.

app/routers/users.py
```diff
@@ -204,10 +204,10 @@ async def update_user(
     This endpoint updates the name field of a user.
     Only the name can be updated.
     """
-    if user.status != UserStatus.ACTIVE:
+    if user.status == UserStatus.DISABLED:
```
ffaraone: It should be DELETED right ?

app/routers/users.py
```diff
@@ -219,8 +219,21 @@ async def update_user(
     dependencies=[Depends(check_operations_account)],
     status_code=status.HTTP_204_NO_CONTENT,
 )
-async def delete_user(id: str):  # pragma: no cover
-    pass  # not yet implemented
+async def delete_user(
```
ffaraone, line 222: We should remove this from here since the right implementation in another PR

ffaraone (approved):

PR #100, opened 2025-02-28
feat: Automate audit fields with sqlalchemy hooks

arturbalabanov:
Automatically set `created_by`, `updated_by`, `created_at`, `updated_at`, `deleted_by` and `deleted_at` without using the handlers.

Why? Not every model creation goes through the handlers in practice (e.g. tests, commands) -- even if we change our habits, there is no way to enforce this and sqlalchemy has built-in mechanisms for this anyway. This is important as otherwise the models are not properly audited and may contain wrong data. Also I was bored :D  As a side effect the handlers look more readable (as they focus on the main business logic) and we can even remove the `soft_delete` method (it's effectively just `self.update({"status": "deleted"})` at this point.

Intentionally leaving this in draft state as it's not complete (a few todos and missing proper tests) -- will wait for feedback before finishing it up.

We can also use hooks like this for selects (e.g. when an affiliate user queries data, `status != 'deleted'` is automatically applied) and this will simplify our codebase even further. But we need to be careful to not affect useful error messages.

### DIFF
--- a/app/db/handlers.py
+++ b/app/db/handlers.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 from collections.abc import AsyncGenerator, Sequence
-from contextlib import suppress
 from datetime import UTC, datetime
 from typing import Any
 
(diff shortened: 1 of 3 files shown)


ffaraone:
I don't get the benefit of such change. In case of commands, there is no actor calling the command and for tests, it can be set explicitly if needed. Anyway, related to deletion of object I prefer to have an explicit delete method in the handler. IMO is much more clear for someone reading this code.

PR #114, opened 2025-03-06
FIX 500 error 

antoniodimariano:
FIX 500 error when calling GET /accounts/{id}/users and DELETE /accounts/{id}/users/{user_id} without authentication


arturbalabanov (approved):
LGTM

ffaraone (approved):

PR #115, opened 2025-03-10
Refactor of Delete user by id

antoniodimariano:
Refactor of Delete user by id
Increase `le` to LimitOffsetParams in pagination.py
Add modelConfig in EmployeeBase

### DIFF
--- a/app/db/handlers.py
+++ b/app/db/handlers.py
@@ -43,10 +43,6 @@ class CannotDeleteError(DatabaseError):
     pass
 
 
-class CannotUpdateStatusError(DatabaseError):
-    pass
-
-
 class ConstraintViolationError(DatabaseError):
     pass
 
--- a/app/db/handlers.py
+++ b/app/db/handlers.py
@@ -6,7 +6,7 @@ from datetime import UTC, datetime
 from typing import Any
 
 import sqlalchemy
-from sqlalchemy import ColumnExpressionArgument, Exists, func, select
+from sqlalchemy import ColumnExpressionArgument, Exists, func, select, update
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import joinedload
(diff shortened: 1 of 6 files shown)


app/db/handlers.py
```diff
@@ -42,6 +43,10 @@ class CannotDeleteError(DatabaseError):
     pass
 
 
+class CannotUpdateStatusError(DatabaseError):
```
ffaraone, line 46: This error is not used anywhere, could yo please remove it ?

ffaraone (approved):

app/db/handlers.py
```diff
@@ -306,6 +316,29 @@ def __init__(self, session):
...
+            .values(
+                status=AccountUserStatus.DELETED,
+                deleted_at=datetime.now(UTC),
+                deleted_by_id=actor_id,
+                # Explicitly set the relationship key
+                # value instead of setting the expected `deleted_at` field. Setting the deleted_by
+                # causes an error due to a misinterpreted query that puts the deleted_at field
+                # expecting a boolean value
+            )
+        )
```
arturbalabanov, line 338: So when we have multiple different ways to update a record we cannot apply automations (in this case since you're not going through the `update` method `updated_by` and `updated_at` won't be set correctly) _unless we intercept the queries with events_ -- that's what I do in #100

So basically I'm using your PR to promote my own ahahah

arturbalabanov: On a more serious note, what's the benefit of doing this refactoring? I can see that it's slightly more efficient (1 query instead of 2) but that's negligible and the downside with your approach is that the user won't receive an error if the update cannot be performed (as the query will always be successful) and that can be worse UX

ffaraone: this is not to save just one single query. A user can have multiple useraccounts it'a a 1 + n.
We are cascading from user to user accounts, I don't understand "if the update cannot be performed (as the query will always be successful) and that can be worse UX"

ffaraone (approved):

app/db/handlers.py
antoniodimariano: as @ffaraone said, we are cascading from `user` to `user account` with just 1 query. Since users can have N accounts, we can update them with one query

PR #116, opened 2025-03-12
Mpt 7011 model handler improvements

antoniodimariano:
initial refactor of the ModelHandler


app/db/handlers.py
```diff
@@ -35,6 +35,10 @@ class DatabaseError(Exception):
     pass
 
 
+class InvalidParameters(Exception):
+    pass
```
arturbalabanov, line 39: Personally I'd remove this error and use `ValueError` instead as that's quite standard for these cases (and is consistent with the rest of the codebase).

But if you'd rather use a custom exception (I don't mind that), here are a couple of suggestions (not mutually exclusive too):

* Change the name to `UsageError`, so that it's more flexible
* Inherit from `ValueError`, so that it still matches a `try: ... except ValueError` clause for backwards compatibility
* SInce that exception is not related to the database maybe move it to `app.utils`?

app/db/handlers.py
```diff
@@ -174,19 +178,101 @@ async def soft_delete(self, id_or_obj: str | M) -> M:
 
         return await self.update(obj, data=column_updates)
 
-    async def fetch_page(
+    def _apply_conditions_to_the_query(
```
arturbalabanov, line 181: Nitpick: can we move all the "private" methods in one place? Currently they are all at the end but I don't mind if you prefer to keep them at the beginning (after the dunder methods like `__init__`)

app/db/handlers.py
```diff
@@ -174,19 +178,101 @@ async def soft_delete(self, id_or_obj: str | M) -> M:
...
-        extra_conditions: Sequence[ColumnExpressionArgument | Exists] | None = None,
-        options: list[ORMOption] | None = None,
-    ) -> Sequence[M]:
-        query = select(self.model_cls).offset(offset).limit(limit).order_by("id")
-        if extra_conditions:
-            query = query.where(*extra_conditions)
+        query: Select,
+        where_clauses: ColumnExpressionArgument | list[ColumnExpressionArgument] | None = None,
+        options: list[ColumnExpressionArgument] | None = None,
+        order_by: list[ColumnExpressionArgument] | str | None = None,
```
arturbalabanov, line 186: In the spirit of [the Zen of Python](https://peps.python.org/pep-0020/):

> There should be one-- and preferably only one --obvious way to do it.

I think we shouldn't allow str to be passed in `order_by` as the standard way to do it in SQLAlchemy would be `Model.column_name`, e.g. `Entitlement.created_at`. This is a bit more verbose than just using `created_at` but it has a few benefits:

* In case of name collision when there are joins (and remember that most of them are hidden from the caller as they are done automatically) it makes it obvious which `created_at` column is used for joining -- "Explicit is better than implicit"
* Since we're not allowing an arbitrary string here, type checking will catch issues like misspelled or missing columns (happens a lot to me :D)
* Similarly, autocompletion will also work, so typing it would be just as fast if not even faster
* If you want your ordering to be descending, you can use: `desc(Entitlement.created_at)` or `Entitlement.created_at.desc()`
* It's just more consistent with the other arguments (I don't know about you but `EntitlementHandler.query_db(where_clauses=[Entitlement.status != 'deleted'], order_by='created')` mildly annoys me :D )

What are your thoughts @ffaraone @antoniodimariano?

app/db/handlers.py
```diff
@@ -174,19 +178,101 @@ async def soft_delete(self, id_or_obj: str | M) -> M:
...
+    ) -> Select:
```
arturbalabanov, line 187: Better to use `Sequence` instead of `list` here (and in the other endpoints) to allow for tuples, sets etc.

app/db/handlers.py
```diff
@@ -174,19 +178,101 @@ async def soft_delete(self, id_or_obj: str | M) -> M:
...
+        """
+        Applies default options and extra conditions to the query.
+
+        Args:
+            query (Select): The query to modify.
+            where_clauses (list[ColumnExpressionArgument] | None): Additional query conditions.
```
arturbalabanov, line 193: You can omit the types here if they are specified in the function signature -- the doc's types will quickly get out of date anyway and will make the documentation misleading.

Btw there must be a linter (in `ruff` or some other tool) for the docstrings, would be great if we enable it (but that's probably out of scope here)

app/db/handlers.py
```diff
@@ -174,19 +178,101 @@ async def soft_delete(self, id_or_obj: str | M) -> M:
...
+                query = query.order_by(*order_by)
+            else:
+                raise TypeError("order_by must be a string or a list of ColumnExpressionArgument.")
         if orm_options:
             query = query.options(*orm_options)
+        return query
+
+    async def query_db(
+        self,
+        where_clauses: Sequence[ColumnExpressionArgument | Exists] | None = None,
```
arturbalabanov, line 217: `Exist` is a subtype of of `ColumnElement`, better to use that for more flexbility

app/db/handlers.py
```diff
@@ -174,19 +178,101 @@ async def soft_delete(self, id_or_obj: str | M) -> M:
...
+        limit: int | None = 50,
+        offset: int = 0,
```
arturbalabanov, line 219: I think the default value of `limit` should be `None` -- this method can be used anywhere, even outside of pagination, so if I want to for example select all deleted entitlements in the last year (to delete them permanently or archive them) I will select only 50 of them (without knowing!).

Remember that `LimitOffsetParams` already makes 50 the default value of `limit`, so that should be enough for pagination.

Similarly, best to default `offset` to `None` -- that wouldn't cause any functional changes but if we do query introspection and check if the query has `offset` it will always return `True`

app/db/handlers.py
```diff
@@ -174,19 +178,101 @@ async def soft_delete(self, id_or_obj: str | M) -> M:
...
+        all_results: bool | None = False,
```
arturbalabanov, line 220: What does `None` value mean here? This should always be bool, no?

app/db/handlers.py
```diff
@@ -174,19 +178,101 @@ async def soft_delete(self, id_or_obj: str | M) -> M:
...
+        order_by: str | None = None,
+        options: list[ORMOption] | None = None,
```
arturbalabanov, line 222: See my previous comment about `order_by`, also we should allow multiple values to be passed to it, right?

And, for both of these, use `Sequence` instead of `list`

app/db/handlers.py
```diff
@@ -174,19 +178,101 @@ async def soft_delete(self, id_or_obj: str | M) -> M:
...
+        query = select(self.model_cls)
+        # add the default options, if any
+        query = self._apply_conditions_to_the_query(
+            query=query, where_clauses=where_clauses, options=options, order_by=order_by
+        )
+        if not all_results:
+            # apply limit and offset
+            query = query.offset(offset).limit(limit)
         results = await self.session.execute(query)
         return results.scalars().unique().all()
```
arturbalabanov, line 277: `unique` should be a parameter, no?

Also, you can use `self.session.scalars(query)` here which is a shorthand for `self.session.execute(query).scalars()` :)

app/db/handlers.py
arturbalabanov: Actually do we even need this? Can't it be implied by `limit = None`?

app/pagination.py
```diff
@@ -47,14 +47,15 @@ async def paginate[M: Base, S: BaseSchema](
...
     items: Sequence[M] = []
     if params.limit > 0:
-        items = await handler.fetch_page(
+        items = await handler.query_db(
             limit=params.limit,
             offset=params.offset,
-            extra_conditions=extra_conditions,
+            where_clauses=where_clauses,
             options=page_options,
+            order_by="id",
```
arturbalabanov, line 58: Does default ordering by `id` make sense with our ID format (specifically when we generate it we just use random numbers)? I think we should either use `created_at` (and make sure that the models have an index on it!), what do you think?

Good spot about the pagination missing ordering btw!

tests/api/test_accounts_api.py
```diff
@@ -84,6 +84,17 @@ async def test_can_create_accounts(
...
+    {"name":"Microsoft",
+    "external_id":"ACC-9044-8753",
+    "created_at":"2025-02-19T09:40:55.850868Z",
+    "updated_at":"2025-02-19T09:40:55.850875Z",
+    "deleted_at":null,"created_by":
+    {"id":"FUSR-5243-4812","type":"user","name":"Antonio Di Mariano"},
+    "updated_by":{"id":"FUSR-5243-4812","type":"user","name":"Antonio Di Mariano"},
+    "deleted_by":null,"id":"FACC-5297-1380","entitlements_stats":null,"status":"active",
+    "type":"affiliate"}
+    """
```
arturbalabanov, line 97: Debug comment you forgot to remove?

arturbalabanov (approved):
Changes look good, I think all my comments are regarding opinions and taste, so by definition not merge blockers -- approving the PR and feel free to resolve them as you see fit.

Overall though I really like the change! :)

app/db/handlers.py
ffaraone: 👍

app/db/handlers.py
ffaraone: Yes, exactly.

app/db/handlers.py
ffaraone: Good idea!

app/pagination.py
ffaraone: We already discussed it with @antoniodimariano and we said to use updated_at descending as the default.

app/db/handlers.py
antoniodimariano: same page here

app/db/handlers.py
antoniodimariano: don't we need a default value for limit?

ffaraone (approved):
Great job 🥇

PR #118, opened 2025-03-12
Replace the old payload of POST /users with a new one without nested objects.

antoniodimariano:
Replace the old payload of POST /users with a new one without nested objects.


ffaraone (approved):
I think we should first agree about having a layer for business logic before doing such refactoring.

antoniodimariano:
> I think we should first agree about having a layer for business logic before doing such refactoring.

Do you mean the refactor of the whole function? I only extracted two methods out of the main function.

PR #123, opened 2025-03-13
MPT-7893 add rql support for /accounts

ffaraone:


app/db/handlers.py
```diff
@@ -268,18 +270,25 @@ async def count(
         Returns:
             int: The count of matching records.
         """
-        query = select(func.count(self.model_cls.id))
+        if base_query is not None:
+            query = select(func.count(self.model_cls.id)).select_from(
+                base_query.get_final_froms()[0]
+            )
+            if base_query.whereclause is not None:
+                query = query.where(base_query.whereclause)
```
arturbalabanov, line 278: I wonder if we can split this argument to `select_from` and add a where clause as an argument? Though if it's too tricky to do with the current design, don't bother, it's not a big deal.

app/routers/accounts.py
```diff
@@ -161,8 +163,11 @@ async def get_account_by_id(
     response_model=LimitOffsetPage[AccountRead],
     dependencies=[Depends(check_operations_account)],
 )
-async def get_accounts(account_repo: AccountRepository):
-    return await paginate(account_repo, AccountRead)
+async def get_accounts(
+    account_repo: AccountRepository,
+    base_query: Select = Depends(RQLQuery(AccountRules())),
```
arturbalabanov, line 168: nitpick / question: are we ever going to supply arguments to the rule class (e.g. use the same rule with different filters being enabled for different endpoints vs creating a seperate classes if that's the case like we do with pydantic models)? If not, then best to pass the class as an argument to `RQLQuery` instead of an instance

app/routers/accounts.py
```diff
@@ -161,8 +163,11 @@ async def get_account_by_id(
...
+):
+    return await paginate(account_repo, AccountRead, base_query=base_query)
```
arturbalabanov, line 170: This looks super clean and I like it already (so no need to make any change). Given that limit and offset is part of the RQL specification though, shouldn't this be handled by `RQLQuery` or the model rule?

app/routers/users.py
```diff
@@ -211,7 +211,7 @@ async def validate_and_get_user(
     if user.status == UserStatus.DISABLED:
         raise HTTPException(
             status_code=status.HTTP_400_BAD_REQUEST,
-            detail=(f"The user {user.email} cannot be invited " "because it is disabled."),
+            detail=(f"The user {user.email} cannot be invited because it is disabled."),
```
arturbalabanov, line 214: Annoying that `ruff` doesn't remove the brackets too :(

app/rql.py
```diff
@@ -0,0 +1,51 @@
...
+
+class ActorRules(ModelRQLRules):
+    __model__ = Actor
+
+    id = FieldRule()
+    name = FieldRule()
+    type = FieldRule()
+
+
+class TimestampMixin:
```
arturbalabanov, line 18: Can we rename this to `TimestampedRule` or something like this? To avoid name collision with the same mixin in the models -- I can see myself automatically importing the wrong one by mistake 😅

app/rql.py
```diff
@@ -0,0 +1,51 @@
...
+        self.rules = rules
+
+    def __call__(self, request: Request) -> Select | None:
+        qs = request.scope["query_string"].decode()
+        parsed = parse_qs(qs, keep_blank_values=True)
+        rql_tokens = [k for k, v in parsed.items() if v == [""]]
+        rql_expression = "&".join(rql_tokens)
+        if not rql_expression:
+            return None
+        return self.rules.build_query(rql_expression)
```
arturbalabanov, line 51: I think this should be part of `requela`, right? Maybe in `compat.fastapi` or `compat.starlette`? Looks like it will always be the same for all models, right?

app/rql.py
```diff
@@ -0,0 +1,51 @@
...
+
+class RQLQuery:
+    def __init__(self, rules: ModelRQLRules):
+        self.rules = rules
+
+    def __call__(self, request: Request) -> Select | None:
+        qs = request.scope["query_string"].decode()
+        parsed = parse_qs(qs, keep_blank_values=True)
+        rql_tokens = [k for k, v in parsed.items() if v == [""]]
+        rql_expression = "&".join(rql_tokens)
```
arturbalabanov, line 48: There must be a method in urllib to do the join too to handle some edgecases, not sure if it's needed here but would be nice to use it

antoniodimariano (approved):

tests/api/test_accounts_api.py
```diff
@@ -761,3 +762,97 @@ async def test_cannot_remove_deleted_user_from_account(
...
+    )
+    await account_factory(
+        name="Second",
+    )
+
+    with capsql:
+        response = await api_client.get(
+            "/accounts?eq(name,First)",
+            headers={"Authorization": f"Bearer {ffc_jwt_token}"},
+        )
```
arturbalabanov, line 784: `capsql` is not doing anything here -- you're not doing any checks with it. But this gave me an idea to add this to the library -- similarly to how `httpx_mock` complains if we're defining mocks which are not being called :)

tests/api/test_accounts_api.py
```diff
@@ -761,3 +762,97 @@ async def test_cannot_remove_deleted_user_from_account(
...
+        name="Milly",
+        status=AccountStatus.ACTIVE,
+        created_by=created_by,
+    )
+
+    with capsql:
+        response = await api_client.get(
+            "/accounts?ilike(events.created.by.name,*racl*)&order_by(-name)",
+            headers={"Authorization": f"Bearer {ffc_jwt_token}"},
+        )
```
arturbalabanov, line 852: Another nitpick, you can use `operations_client` here instead of passing the authentication header manually).

On a more serious note, better to change the `order_by` to be `name` instead of `-name` to make sure the ordering was applied -- `Pilly` was created before `Milly`, so the same response will be returned without `order_by` (if the default ordering is by `updated_by` as we discussed)

arturbalabanov (approved):
Wow this looks super clean, great work @ffaraone! Left a few comments but they are all stylistic, so resolve as you see fit and merge it after

app/rql.py
ffaraone: It will be part of contrib.fastapi for sure. It is on the roadmap but I need to have it integrated to test it a bit more extensively.

app/rql.py
ffaraone: There's `urljoin` but it joins fragments with `/`.  According to the rql grammar, there shouldn't be such edge cases but I will doublecheck.

app/rql.py
ffaraone: I like that the name is exactly the same as for models because it make it clear which fields this mixin has. To avoid your automagic tools importing from the wrong place, probably we can split the rql.py in module per model and have a base/core with such common stuff. In the meanwhile don't worry, your tests will fail if you import the wrong one ;)

tests/api/test_accounts_api.py
ffaraone: Yeah, I have in my personal backlog of tech debts, plenty of tests for checking several combinations of filters.

app/routers/accounts.py
arturbalabanov: > especially because in the SWO syntax there is not operator for that but they are passed as usual querystring parameters

Ah that makes sense, sure let's leave it as it is then :)

PR #129, opened 2025-03-19
Add RQL to Organization Model.

antoniodimariano:
Add RQL to Organization Model.
FIX bug when filtering by datetime

### DIFF
--- a/app/rql.py
+++ b/app/rql.py
@@ -1,10 +1,10 @@
-from urllib.parse import parse_qs
+from urllib.parse import parse_qs, quote
 
 from fastapi import Request
 from requela import FieldRule, ModelRQLRules, RelationshipRule
 from sqlalchemy.sql.selectable import Select
 
-from app.db.models import Account, Actor
+from app.db.models import Account, Actor, Organization
 
 
 class ActorRules(ModelRQLRules):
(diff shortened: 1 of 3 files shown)


tests/api/test_organizations_api.py
```diff
@@ -24,6 +24,136 @@ async def test_get_all_organizations_empty_db(api_client: AsyncClient, ffc_jwt_t
...
+
+
+async def test_get_organization_with_not_valid_filter(
+    organization_factory: ModelFactory[Organization], api_client: AsyncClient, ffc_jwt_token: str
+):
+    await organization_factory(operations_external_id="EXTERNAL_ID_1")
+    with pytest.raises(ValueError):
+        await api_client.get(
+            "/organizations?eq(ciaociao,active)",
+            headers={"Authorization": f"Bearer {ffc_jwt_token}"},
```
arturbalabanov, line 153: No need for a change just FYI -- you can use `operations_client` here (there is a similar `affiliate_client` too) and make the test shorter: `await operations_client.get("/organizations?eq(ciaociao,active)")`

app/rql.py
```diff
@@ -37,12 +37,24 @@ class AccountRules(ModelRQLRules, AuditableMixin):
...
+
 class RQLQuery:
     def __init__(self, rules: ModelRQLRules):
         self.rules = rules
 
     def __call__(self, request: Request) -> Select | None:
-        qs = request.scope["query_string"].decode()
+        qs = quote(
+            request.scope["query_string"].decode(), safe="/&()=_.-~:,"
+        )  # make sure we can decode datetime
```
arturbalabanov, line 57: Good spot!

arturbalabanov (approved):

tests/api/test_organizations_api.py
antoniodimariano: a `ValueError` will be raised even with the `operation_client`

PR #131, opened 2025-03-20
Add RQL to System Model.

antoniodimariano:


app/rql.py
```diff
@@ -27,6 +27,15 @@ class AuditableMixin(TimestampMixin):
     deleted_by = RelationshipRule(alias="events.deleted.by", rules=ActorRules())
 
 
+class SystemRules(ModelRQLRules, AuditableMixin):
+    __model__ = System
+
+    id = FieldRule()
+    external_id = FieldRule()
+    owner_id = FieldRule()
```
ffaraone, line 35: This should be:

`owner = RelationshipRule(rules=AccountRules())`

ffaraone (changes requested):
Please check other rules classes for relationships. For example in Entitlements we should have

`redeemed_by = RelationshipRule(alias="events.redeemed.by", rules=OrganizationRules())`

app/rql.py
```diff
@@ -27,22 +27,22 @@ class AuditableMixin(TimestampMixin):
...
-class AccountRules(ModelRQLRules, AuditableMixin):
-    __model__ = Account
+class SystemRules(ModelRQLRules, AuditableMixin):
+    __model__ = System
 
     id = FieldRule()
-    name = FieldRule()
     external_id = FieldRule()
-    type = FieldRule()
+    owner_id = RelationshipRule(rules=AccountRules(), alias="owner_id")
```
ffaraone, line 45: this should be just owner and no need to add an alias. If you want to filter by owner id you should use for example `eq(owner.id,FACC-1234-5678)`

app/rql.py
```diff
@@ -56,6 +56,16 @@ class OrganizationRules(ModelRQLRules, AuditableMixin):
...
 
 
+class EntitlementRules(ModelRQLRules, AuditableMixin):
+    __model__ = Entitlement
+
+    id = FieldRule()
+    name = FieldRule()
+    datasource_id = FieldRule()
+    status = FieldRule()
+    redeemed_by = RelationshipRule(alias="redeemed_by", rules=OrganizationRules())
```
ffaraone, line 66: the alias must be: `events.redeemed.by`

ffaraone (changes requested):

app/rql.py
antoniodimariano: is this one correct @ffaraone 
```
response = await api_client.get(
        f"/entitlements?eq(events.redeemed.by,{entitlement_gcp.redeemed_by_id})",
        headers={"Authorization": f"Bearer {gcp_jwt_token}"},
    )
```
Because I get this error 


```
if not relations:
>           raise ValueError(f"Relation with alias '{alias}' not found")
E           ValueError: Relation with alias '' not found

../../.venv/lib/python3.12/site-packages/requela/rules.py:203: ValueError
```

app/rql.py
```diff
@@ -27,22 +27,22 @@ class AuditableMixin(TimestampMixin):
...
-class AccountRules(ModelRQLRules, AuditableMixin):
-    __model__ = Account
+class SystemRules(ModelRQLRules, AuditableMixin):
+    __model__ = System
 
     id = FieldRule()
-    name = FieldRule()
     external_id = FieldRule()
-    type = FieldRule()
+    owner = RelationshipRule(rules=AccountRules(), alias="owner")
```
ffaraone, line 45: The alias is not needed if it is the same as the field name.

app/rql.py
```diff
@@ -56,6 +56,17 @@ class OrganizationRules(ModelRQLRules, AuditableMixin):
...
 
+class EntitlementRules(ModelRQLRules, AuditableMixin):
+    __model__ = Entitlement
+
+    id = FieldRule()
+    name = FieldRule()
+    datasource_id = FieldRule()
+    status = FieldRule()
+    redeemed_by = RelationshipRule(rules=OrganizationRules(), alias="events.redeemed.by")
+    terminated_by = RelationshipRule(rules=ActorRules(), alias="events.terminated.by")
```
ffaraone, line 67: Good addition! Could you please add also `terminated_at = FieldRule(alias="events.terminated.at")` ?

ffaraone (approved):

PR #132, opened 2025-03-21
Add RQL to User model

antoniodimariano:


app/rql.py
```diff
@@ -27,6 +27,19 @@ class AuditableMixin(TimestampMixin):
...
+    __model__ = AccountUser
+
+
+class UserRules(ModelRQLRules, AuditableMixin):
+    __model__ = User
+
+    id = FieldRule()
+    email = FieldRule()
+    status = FieldRule()
+    accounts = RelationshipRule(rules=AccountUserRules(), alias="accounts")
```
ffaraone, line 40: I don't understand what we can filter with this. It will be awesome to be able to issue queries like:

`any(accounts, eq(accounts.account.id,FACC-1234-5678))` but I'm not sure requela already support it.

PR #133, opened 2025-03-24
feat: Add a command to fetch and store into the db monthly datasource costs (MPT-7851)

arturbalabanov:
* [x] Add a new model to store the datasource costs
* [x] Add migrations for the new model
* [x] Add a command to fetch the datasource costs from optscale and store them in the database
* [x] Add a bulk upsert function to optimise the number of database queries we're doing
* [x] Plenty of tests


app/db/base.py
```diff
@@ -21,11 +21,16 @@ def get_db_engine(settings: AppSettings) -> AsyncEngine:
...
 async def get_db_session(
     db_engine: AsyncEngine = Depends(get_db_engine),
 ) -> AsyncGenerator[AsyncSession]:
-    async_session = async_sessionmaker(bind=db_engine, class_=AsyncSession, expire_on_commit=False)
-    async with async_session() as session:
+    AsyncSession = get_sessionmaker(db_engine)
+
+    async with AsyncSession() as session:
         async with session.begin():
             yield session
```
arturbalabanov, line 35: Necessary to decouple creating a session without starting a transaction while doing minimal changes to `app.db`. Still far from ideal, I have a larger refactoring PR coming after this to address some of the issues.

app/commands/update_last_month_datasource_expenses.py
```diff
@@ -0,0 +1,139 @@
...
+    return datasources_per_organization_id
+
+
+async def store_datasource_expenses(
+    datasource_expense_handler: DatasourceExpenseHandler,
+    datasources_per_organization_id: dict[str, list[dict]],
+    year: int,
+    month: int,
+) -> None:
+    expenses_to_upsert = [
```
ffaraone, line 71: We expect an organization to have few data sources (<10), so we most likely won't get much performance gain from doing this bulk upsert. With that in mind, do you think we can follow a different approach to simplify the code and gain in readability ?

app/commands/update_last_month_datasource_expenses.py
```diff
@@ -0,0 +1,139 @@
...
+
+    async with sessionmaker() as session:
+        datasource_expense_handler = DatasourceExpenseHandler(session)
+        organization_handler = OrganizationHandler(session)
+
+        async with session.begin():
+            logger.info("Querying organizations with no recent datasource expenses")
+            organizations = await organization_handler.query_db(
+                where_clauses=[
+                    ~Organization.datasource_expenses.any(DatasourceExpense.updated_at >= today)
```
ffaraone, line 102: If I understood, do you want to get only organization where at least one datasource was not updated today right ? in that case, is it a bit clearer to express the query like 
`Organization.datasource_expenses.any(DatasourceExpense.updated_at < today)` ?

ffaraone (changes requested):

app/commands/__init__.py
```diff
@@ -5,6 +5,7 @@
     invite_user,
     openapi,
     redeem_entitlements,
+    update_last_month_datasource_expenses,
```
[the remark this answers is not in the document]
arturbalabanov: Changed

app/commands/update_last_month_datasource_expenses.py
arturbalabanov: You're right -- it was a premature optimisation on my part. For some reason I thought each organization could have potentially thousands of datasources but that doesn't really make sense 😆 Rolled it back to the initial version, it generates more queries but shouldn't be an issue and if it is, we already know how to optimise it :)

PR #135, opened 2025-03-25
Add RQL to ChargesFile model

antoniodimariano:
Add RQL to ChargesFile model
Add GET /charges endpoint
Add Migration to replace Numeric precision from 10 to 18 digits
Replace wrong base_query Model in get_user_accounts
Update Requela version to 0.1.11 to support decimal fields


app/routers/chargesfiles.py
```diff
@@ -10,9 +13,15 @@
     "",
     response_model=LimitOffsetPage[ChargesFileRead],
 )
-async def get_charges_files():
-    # pending implementation
-    pass  # pragma: no cover
+async def get_charges_files(
```
ffaraone, line 16: We must force the logic so affiliates can only see charges file they own like we do with systems https://github.com/softwareone-platform/ffc-finops-operations/blob/main/app/routers/systems.py#L21.

ffaraone (changes requested):

ffaraone (approved):

PR #136, opened 2025-03-26
feat: Add a command to fetch the latest currency conversion rates and store them in the DB (MPT-8063)

arturbalabanov:
I definitely went overboard with the changes here 😅 

Also includes some refactoring to the mock clients, little by little testing external APIs will become easier and easier :)

Also, intentionally didn't include the currency conversion itself in this PR as it's quite large as it is. Will open a separate ticket for that


app/db/models.py
```diff
@@ -305,6 +315,48 @@ class DatasourceExpense(Base, HumanReadablePKMixin, AuditableMixin):
     )
 
 
+class ExchangeRates(Base, HumanReadablePKMixin, AuditableMixin):
```
ffaraone, line 318: IMO it should inherit from TimestampMixin, who creates, updates this object is always the command under cron schedule, so the *_by will always be null.

ffaraone (changes requested):

app/api_clients/exchange_rate.py
```diff
@@ -0,0 +1,75 @@
...
+    # Exchange Rate API can return responses with a "result" field that is not "success"
+    # while still being 200 OK responses
+    #
+    # ref: https://www.exchangerate-api.com/docs/standard-requests
+    response: httpx.Response
+    response_result: str
+    error_type: ExchangeRateAPIErrorType
+
+    def __init__(self, response: httpx.Response):
+        data = response.json()
```
antoniodimariano, line 34: I would wrap the `response.json()` in a try-except to handle invalid JSON.

app/api_clients/exchange_rate.py
```diff
@@ -0,0 +1,75 @@
...
+            return exc.should_retry()
+
+        # Retry on all other httpx errors (e.g. timeouts)
+        return isinstance(exc, httpx.HTTPError)
+
+    @stamina.retry(on=retry_condition, attempts=3)
+    async def get_latest_rates(self, base_currency: str) -> httpx.Response:
+        response = await self.httpx_client.get(f"/latest/{base_currency}")
+        response.raise_for_status()
+        response_data = response.json()
```
antoniodimariano, line 70: same here, like 
```
try:
            response_data = response.json()
        except ValueError:
            raise ExchangeRateAPIError(response)
```

app/api_clients/exchange_rate.py
```diff
@@ -0,0 +1,75 @@
...
+
+        if response_data["result"] != "success":
```
antoniodimariano, line 72: Is there here a potential KeyError, or `result` is always in the `response_data`?

antoniodimariano (approved):
LGTM

app/db/models.py
ffaraone: 👍

app/api_clients/exchange_rate.py
```diff
@@ -0,0 +1,83 @@
...
+    response: httpx.Response
+    response_result: str | None
+    error_type: ExchangeRateAPIErrorType
+
+    def __init__(self, response: httpx.Response):
+        self.response = response
+
+        try:
+            data = response.json()
+            self.response_result = data.get("result")
```
antoniodimariano, line 38: `response_result` can be `None` here

arturbalabanov: yep, same if `ValueError` occurs, even changed the type annotation above, this is the desired behaviour

ffaraone (approved):

PR #137, opened 2025-03-27
chore: Make running the tests A LOT faster

arturbalabanov:
**As a result of this PR _all tests_ now run in 22s on my machine. Without these changes they run in 184s, so that more than 8x improvement :)**

-------

This is a fairly large PR, so let me break it down:

The most important changes are in `conftest.py`:

1. Build the database schema only at the beginning of the test session, not before each test. This is a huge performance improvement, as building the schema is the most time-consuming part of the test suite.
2. To keep the tests isolated, each test is run in a transaction which always rolls back at the end of the test. Any transactions opened by the application are replaced with nested transactions. This is the [recommended approach by SQLAlchemy](https://docs.sqlalchemy.org/en/20/orm/session_transaction.html#joining-a-session-into-an-external-transaction-such-as-for-test-suites) and once set up seems to work perfectly for our codebase.

This change required quite a lot of restructuring of the DB sessions code and while I was at it I did one more optimisation: I noticed that we're creating a new DB engine for every request (part of the FastAPI dependancies) which creates its own session pool, defeating the purpose of it in the first place. So, I made the DB engine and the sessionmaker global (again, following [the SQLAlchemy recommendations](https://docs.sqlalchemy.org/en/20/orm/session_basics.html#when-do-i-make-a-sessionmaker)) and create a session when needed. SQLAlchemy works best that way as sessions will share connections from the pool as needed and this will allow us to support substaintially more concurrent requests. Take a look at [this talk](https://www.youtube.com/watch?v=Ehhq5oYbbuY) for more detailed explanation. We still need to do one more change (only call `session.begin()` as needed) to fully take advantage of this but that would require even more changes, so we can address it later in a seperate PR as we need it.

The other changes are mostly just restructuring of the dependancies to avoid circular imports (and that way the dependancies are more discoverable) and changes to the commands to move the DB setup in a single place.


app/auth/auth.py
arturbalabanov, line 1: The changes in this file are just moving the dependancies to `app/dependancies.py`, I didn't make any changes to the functions themselves

app/commands/redeem_entitlements.py
```diff
@@ -20,20 +19,16 @@
...
     return response.json()["cloud_accounts"]
 
 
-async def redeem_entitlements(
-    settings: Settings,
-):
-    engine = get_db_engine(settings)
-    async with asynccontextmanager(get_db_session)(engine) as session:
+async def redeem_entitlements(settings: Settings):
+    # FIXME: Long-lived DB transaction (making API calls inside the transaction)
```
arturbalabanov, line 29: This is a potential issue as we're making a lot of API calls inside the transaction (it's what we were doing before this change too btw) but it's not easy to change as the API calls are in the `stream_scalars` `for` loop block. It's not a huge issue though as this is just a CLI command, so left this `FIXME` and we can address it as needed

app/dependencies.py
arturbalabanov, line 1: This file grew quite large and with the different sections it makes perfect sense to split into submodules. Only reason I haven't done it is to keep the changes as low as possible in this PR, let's do that after this is merged

tests/commands/test_invite_user.py
```diff
@@ -79,12 +83,14 @@ async def test_invite_user_already_invited_force(
...
+
     accountuser_handler = AccountUserHandler(db_session)
     db_account_user = await accountuser_handler.get_account_user(operations_account.id, user.id)
+
     assert db_account_user is not None
     assert db_account_user.status == AccountUserStatus.INVITED
-    assert db_account_user.invitation_token != account_user.invitation_token
-    assert db_account_user.invitation_token_expires_at != account_user.invitation_token_expires_at
+    assert db_account_user.invitation_token != original_invitation_token
+    assert db_account_user.invitation_token_expires_at != original_invitation_token_expires_at
```
arturbalabanov, line 93: These changes are expected due to SQLAlchemy's session automatically updated `account_user` when the model changed. I think this wasn't needed before because the tests and the API were creating their own sessions and using different connection pools

ffaraone (approved):

ffaraone:
Great job 🥇

PR #138, opened 2025-03-27
Mpt 8354 upload zipped charges file to azure storage blob

antoniodimariano:
Add Functionalities to upload charges file to Azure Blob Storage 
Add Azurite to docker-compose and github actions


app/blob_storage.py
```diff
@@ -0,0 +1,70 @@
+import re
+from functools import lru_cache
+
+from app.api_clients.azure import AzureBlobServiceClient
+from app.conf import get_settings
+
+
+@lru_cache
+def get_azure_blob_service_client() -> AzureBlobServiceClient:
```
ffaraone, line 9: A function that merely instantiates a class and does nothing else is typically unnecessary because it introduces an extra layer of function call overhead without adding any real value.

Furthermore, decorating this function with lru_cache is counterproductive because:

1) The function always receives the same argument so the cache will store only one instance of the class.
2) creating an instance of Azure BlobServiceClient is not an expensive operation, then caching it does not provide a meaningful performance improvement. Instead, it just adds the overhead of maintaining the LRU cache.

app/api_clients/azure.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class AzureBlobServiceClient:
+    def __init__(
+        self,
+        container_name: str,
+        connection_string: str | None = None,
+    ):
+        if connection_string:
+            self.blob_service_client = BlobServiceClient(
```
ffaraone, line 26: you should use the async version of such client (https://learn.microsoft.com/es-es/python/api/azure-storage-blob/azure.storage.blob.aio.blobserviceclient?view=azure-python)

app/api_clients/azure.py
```diff
@@ -0,0 +1,85 @@
...
+                account_url=connection_string, credential=AZURE_SA_CREDENTIALS
+            )
+        else:  # pragma: no branch
+            raise ValueError("The connection_string must be provided.")
+        self.container_name = container_name
+
+    def get_or_create_container_name(self):
```
ffaraone, line 33: You should convert this method to async using the async client.

app/api_clients/azure.py
```diff
@@ -0,0 +1,85 @@
...
+        """
+        # Create container if not exists
+        try:
+            container_client = self.blob_service_client.create_container(self.container_name)
+        except ResourceExistsError:
+            logger.debug("The Azure storage container already exists. No panic!")
+            container_client = self.blob_service_client.get_container_client(self.container_name)
+        return container_client
+
+    def upload_file_to_azure_blob(
```
ffaraone, line 45: same here. furthermore, we should use https://pypi.org/project/aiofiles/ here.

app/blob_storage.py
```diff
@@ -0,0 +1,70 @@
...
+def validate_year_and_month_format(month: int, year: int):
+    """
+    This function ensures that month and year have a valid format.
+    Year: from 2025 -> 2999
+    Month: cannot be negative, greater than 12.
+    A '0' is added as a prefix if month has 1 digit and valid format.
+    """
+    if month <= 0 or month > 12:
+        raise ValueError("Invalid month format.")
+    month = int(month)
```
ffaraone, line 63: it should already be an int, the function is typed and you already compared in line 61.

app/blob_storage.py
```diff
@@ -0,0 +1,70 @@
...
+    if len(str(abs(month))) == 1:
```
ffaraone, line 64: the following two lines can be simplified with:

```python
month = f"{month:02}"
```

app/blob_storage.py
```diff
@@ -0,0 +1,70 @@
...
+        month = "0" + str(int(month))
+    # validate years starting from 2025 to 2999
+    pattern = re.compile(r"^20(2[5-9]|[3-9][0-9])$|^2[1-9][0-9]{2}$")
```
ffaraone, line 67: pattern matching is a costly operation. why don't check it like an integer greater then 1970 ?

ffaraone (changes requested):

app/blob_storage.py
antoniodimariano: True.

ffaraone (approved):

PR #147, opened 2025-03-31
Add download charges file from Azure Blob Storage

antoniodimariano:
Add download charges file from Azure Blob Storage


app/api_clients/azure.py
```diff
@@ -86,3 +88,42 @@ async def upload_file_to_azure_blob(
         except Exception as error:  # pragma: no branch
             logger.error(f"Unexpected error occurred: {error}")
         return None
+
+    async def download_file_from_azure_blob(
+        self, blob_name: str, account_key: str, sas_expiration_token_min: int
```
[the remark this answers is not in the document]
antoniodimariano: I would move `account_key`  and also the `sas_expiration_token_min` to the constructor but leave the `blob_name` as a parameter of the function because

app/api_clients/azure.py
```diff
@@ -86,3 +88,42 @@ async def upload_file_to_azure_blob(
...
+               Download a blob from Azure Blob Storage using a time-limited SAS token.
+               Please note that the function generate_blob_sas() does not raise an exception
+               even if you pass an expiry time in the past.
+               It just returns a SAS token that is already expired and won't work when used.
+
+               p.s.: If you're using Azurite, make sure you're using the local emulator URL
+               with the following format
+               http://127.0.0.1:10000/devstoreaccount1/your-container/your-blob
+               to download the file, instead of https://devstoreaccount1.blob.core.windows.net/your-container/your-blob
+        http://127.0.0.1:10000/devstoreaccount1/ffc-charges-files/EUR/2025/03/FCHG-1234-5678-9012.zip?se=2025-03-31T17%3A12%3A40Z&sp=r&sv=2025-05-05&sr=b&sig=OuMkcr0teQmIhFqoGPmLK/NPFX%2BRSenXJ7PkAv1v%2BEg%3D
```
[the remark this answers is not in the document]
antoniodimariano: yes, but I guess it's not needed

app/api_clients/azure.py
```diff
@@ -19,10 +21,14 @@ def __init__(
         self,
         account_url: str,
         container_name: str,
+        account_key: str,
         max_block_size: int = 1024 * 1024 * 4,
         max_single_put_size: int = 1024 * 8,
         max_concurrency: int = 4,
+        sas_expiration_token_mins: int = 5,
     ):
+        self.account_key = account_key  # the primary access key of the Azure Storage account.
```
ffaraone, line 30: For the future, it will be much more clean and clear if we set the members in the same order they appear into the constructor, unless there is a good reason to not do that.

ffaraone (changes requested):

tests/azure_blob_storage/test_azure_upload.py
```diff
@@ -30,6 +32,20 @@ async def test_cannot_upload_file():
...
+
+    response = await download_changes_file(
+        file_path=zip_file_path, currency="eur", year=2025, month=3
+    )
+    assert response is not None
+    assert isinstance(response, str)
+    url_parsed = urlparse(response)
+    path_parts = url_parsed.path.lstrip("/").split("/")
+    blob_name = "/".join(path_parts[1:])
+    assert blob_name == "EUR/2025/03/FCHG-1234-5678-9012.zip"
```
arturbalabanov, line 46: Since you're running `datetime.now(UTC)` in one of the functions, this test (and potentially the others in this file) can be flaky (same issue as the one we fixed in `test_update_current_month_datasource_expenses.py`), so worth adding `time_machine.travel` decorators to avoid that

app/api_clients/azure.py
arturbalabanov: I think it makes sense to pass the settings as the only argument in the `__init__` just like we do with the other API clients and this will simplify the usage of the client itself.

To be clear, I think the current design is actually better as ideally the clients shouldn't be coupled to the app's settings but this is the already established pattern for the other clients and it's better to stick to it.

No need for a change now, we can do that later

app/conf.py
```diff
@@ -40,9 +40,12 @@ class Settings(BaseSettings):
 
     azure_sa_container_name: str = "ffc-charges-files"
     azure_sa_url: str
+    azure_sa_account_key: str
+    azure_sas_token_expiration_min: int = 5
```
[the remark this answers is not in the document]
antoniodimariano: ops..

tests/azure_blob_storage/test_azure_upload.py
ffaraone: @arturbalabanov, but I think Azurite will check the validity of the SAS token, so you cannot use `time_machine.travel`. You have to assume flakiness but with 5 minutes validity by default it should fail near never.

arturbalabanov: Ah, yes, fair enough, thanks

app/api_clients/azure.py
antoniodimariano: Yes, we are on the same page here. I see the clients not coupled to the app's settings and free to be initialized through init parameters, making the context more evident.  My 2 cents: We should probably refactor the other API clients to initiate them without passing the settings as the only arg.

arturbalabanov (approved):

ffaraone (approved):

PR #158, opened 2025-04-07
Add to ignore files.

jentyk:
Adding few more exclusions to .dockerignore. It should further reduce docker build time, especially in local environment of developers.
Also, added .idea dir to .gitignore.


arturbalabanov (approved):
Looks good, thank you for making the change :)

PR #159, opened 2025-04-07
feat: Add a command generate charge files every month (MPT-8064)

arturbalabanov:


app/commands/generate_monthly_charges.py
```diff
@@ -0,0 +1,295 @@
...
+        settings = get_settings()
+
+        if exp.organization.linked_organization_id is None:
+            raise ValueError(
+                f"Cannot generate charge for datasource expense {exp.id}: "
+                f"Organization {exp.organization.id} does not have a linked organization ID."
+            )
+
+        month_start = datetime(exp.year, exp.month, 1)
+        month_end = month_start + relativedelta(day=31)
```
arturbalabanov, line 55: Change that to up to last time we updated the data source expense (and also make sure that the datasource expences' updated_at field to alaways be updated)

app/commands/generate_monthly_charges.py
```diff
@@ -0,0 +1,313 @@
...
+            subscription_search_value=exp.organization_id,
+            item_search_criteria="item.externalIds.vendor",
+            item_search_value=settings.ffc_external_product_id,
+            usage_start_time=usage_start,
+            usage_end_time=usage_end,
+            price=price,
+            external_reference=exp.organization.linked_organization_id,
+            vendor_description_1=exp.datasource_name,
+            vendor_description_2=exp.datasource_id,
+            vendor_reference="",
```
antoniodimariano, line 76: Do we need quantity as well?

app/commands/generate_monthly_charges.py
```diff
@@ -0,0 +1,313 @@
...
+        all_orgs_expenses: list[DatasourceExpense] = []
+
+        for organization in organizations:
+            logger.info(
+                "Querying datasource expenses for organization %s for month = %s, year = %s",
+                organization.id,
+                last_month.month,
+                last_month.year,
+            )
+            org_expenses = await datasource_expense_handler.query_db(
```
antoniodimariano, line 130: if we have 10 organizations, we will make 10 separate queries. Not sure if it might be the case of using `.in_()` to run just one query to retrieve all organizations for the given month and year

app/commands/generate_monthly_charges.py
```diff
@@ -0,0 +1,313 @@
...
+        )
+        logger.info("Found %d organizations to process", len(organizations))
+
+        all_orgs_expenses: list[DatasourceExpense] = []
+
+        for organization in organizations:
+            logger.info(
+                "Querying datasource expenses for organization %s for month = %s, year = %s",
+                organization.id,
+                last_month.month,
```
antoniodimariano, line 127: It would probably be better to extract `month, year = last_month.month, last_month.year` instead of referencing them multiple times.

app/commands/generate_monthly_charges.py
arturbalabanov: Good idea, changed :)

app/commands/generate_monthly_charges.py
arturbalabanov: Honestly, I prefer it this way as it's more clear it refers to last month's date and may avoid any confusion

app/commands/update_current_month_datasource_expenses.py
```diff
@@ -96,6 +96,7 @@ async def store_datasource_expenses(
                 month=month,
                 defaults={
                     "month_expenses": datasource["details"]["cost"],
```
antoniodimariano, line 98: Is it possible that those keys are not defined and cause an error?

antoniodimariano (approved):
LGTM

app/commands/update_current_month_datasource_expenses.py
arturbalabanov: In theory no, the optscale docs say this field (and the others we use in the script) will always be present. Of course, in practice the API response may be different but given that this value is the whole reason we're running this script, it should fail if the field is missing. So I'll leave it as it is and if it causes any problems down the line we can go back to it

PR #173, opened 2025-04-24
feat: Fetch exchange rates for multiple base currencies (MPT-8606)

arturbalabanov:
This is the first PR covering only fetching rates for other base currencies to make it easier to review. In future PRs I'll cover changing the conversion logic to take into account the other base currency exchange rates and exporting the correct Exchangerates API response when generating charges files

### DIFF
--- a/app/commands/generate_monthly_charges.py
+++ b/app/commands/generate_monthly_charges.py
@@ -493,7 +493,7 @@ async def main(exports_dir: pathlib.Path, settings: Settings) -> None:
     async with session_factory() as session:
         async with session.begin():
             unique_billing_currencies = await fetch_unique_billing_currencies(session)
-            currency_converter = await CurrencyConverter.from_db(session)
+            currency_converter = await CurrencyConverter.from_db(session, "USD")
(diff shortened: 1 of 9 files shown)


app/commands/generate_monthly_charges.py
```diff
@@ -493,7 +493,7 @@ async def main(exports_dir: pathlib.Path, settings: Settings) -> None:
     async with session_factory() as session:
         async with session.begin():
             unique_billing_currencies = await fetch_unique_billing_currencies(session)
-            currency_converter = await CurrencyConverter.from_db(session)
+            currency_converter = await CurrencyConverter.from_db(session, "USD")
```
arturbalabanov, line 496: This is for backwards compatibility and to make this PR as small as possible -- in future PRs I'll change this to fetch the latest exchange rates for all currencies in the database

ffaraone: Uhmm I'm not sure we should change this, it can happens that there is a new organization with a currency for which no exchange rate has been fetched (because the job didn't run yet) and this is a way to detect such situation

PR #179, opened 2025-05-02
feat: make generation of monthly charges more memory effecient (MPT-9570)

arturbalabanov:
After many headaches with pandas I gave up on using it since using it in append mode is just a wrapper around openpyxl (other xlsx engines don't support append mode) but with more annoying interface.

With the `write_only` mode openpyxls guarantees max 10MB memory usage for _any_ file. In addition, we're now streaming the datasource expenses instead of holding the full list in memory, so the whole script is very scalable memory-wise.

With that being said, we're still generating the files one account at a time and re-fetching the datasource expenses for each of them (even though for the same currency the results will be the same). I decided to go with this approach because:

1. We don't have big restrictions for the runtime of the script, so even if it's a bit slow that should be fine and want to avoid premature optimisations
2. The code is simpler and easier to follow this way. Otherwise we need to keep track of multiple files (some of them potentially empty, extract the upload at the very end (so that it's not done in a db transaction) and similar workarounds for the charge file db records.
3. With the current approach it will be significantly easier to implement the other additions we're planning for the script


app/commands/generate_monthly_charges.py
```diff
@@ -95,18 +95,43 @@ class ChargesFileGenerator:
...
-        return [
-            charge_entry
-            for ds_exp in datasource_expenses
-            for charge_entry in self.charge_entries_for_datasource_expense(ds_exp)
-        ]
-
-    def charge_entries_for_datasource_expense(
-        self, datasource_expense: DatasourceExpense
-    ) -> list[ChargeEntry]:
+    def __post_init__(self) -> None:
```
ffaraone, line 98: I would strongly suggest to convert this from a dataclass to a class. 
Dataclasses are intended for when you primarily need to store data.

app/commands/generate_monthly_charges.py
```diff
@@ -498,39 +455,31 @@ async def main(exports_dir: pathlib.Path, settings: Settings) -> None:
...
+                    if generated_charges_file is not None:
+                        continue
 
-                async with session.begin():
-                    datasource_expenses = await fetch_datasource_expenses(
-                        session, account, currency
+                    generator = ChargesFileGenerator(
+                        account, currency, currency_converter, exports_dir
                     )
+                    async for ds_exp in fetch_datasource_expenses(session, currency):
```
ffaraone, line 469: This fetch should be moved outside the loop of accounts. otherwise you will fetch the expenses for the same currency N times where N is the amount of accounts.

app/commands/generate_monthly_charges.py
```diff
@@ -498,39 +455,31 @@ async def main(exports_dir: pathlib.Path, settings: Settings) -> None:
...
 
-                async with session.begin():
                     charges_file_db_record = await get_or_create_draft_charges_file(
                         session, account, currency, today
                     )
 
-                exported_filepath = charges_file_generator.export_to_zip(
-                    df, filename=f"{charges_file_db_record.id}.zip"
-                )
+                zip_file_path = generator.make_archive(f"{charges_file_db_record.id}.zip")
```
ffaraone, line 479: This should be run into a thread otherwise you will block the event loop.

ffaraone (changes requested):

app/commands/generate_monthly_charges.py
arturbalabanov: Good point! Done

app/commands/generate_monthly_charges.py
arturbalabanov: Is that an issue in this scenario though? I'm asking because:

1. I _think_ making the archive will be a fairly quick operation anyway, so it may not be a big deal
2. We don't have large performance requirements, so even if some operations are done sequentially it will be fine
3. The file I/O of generating the excel files is synchronous anyway and I don't think `openpyxls` supports `asyncio`. We could use threads for that too but it would be an overkill I think

On the other hand, adding threads, making sure they are in sync, handling exceptions inside them etc. just introduces a lot more complexity. I don't think it's worth it but maybe I'm missing something?

app/commands/generate_monthly_charges.py
arturbalabanov: Yep, initially that's what I tried and it's possible but as I wrote in the description it will increase the code complexity a lot. The main reason for this is that this for loop needs to run in a transaction that's only closed when the loop ends. That means that the archive generation and (maybe most importantly) the file upload to Azure will block the transaction unless a large refactoring is done. The excel file generation is even now part of that transaction but we don't really have a choice and it shouldn't take too long anyway. But that's only for one file.

We could extract these operations after the loop but I honestly don't think it's worth it as the code becomes very hard to follow. At the moment making the script more flexible is fairly easy but it will be harder if we do this. Feels like premature optimisation to me, shall we leave it as it is and go back to it if it becomes a problem?

app/commands/generate_monthly_charges.py
```diff
@@ -222,10 +199,11 @@ def export_to_zip(self, df: pd.DataFrame, filename: str) -> pathlib.Path:
...
-        excel_filepath = self.export_to_excel(df, filename="charges.xlsx")
+        excel_filepath = self.save("charges.xlsx")
+        filepath = self.exports_dir / filename
 
         with zipfile.ZipFile(filepath, mode="w") as archive:
-            archive.write(excel_filepath, arcname=excel_filepath.name)
+            archive.write(excel_filepath, arcname="charges.xlsx")
             archive.writestr(
                 f"exchange_rates_{self.currency}.json",
                 self.currency_converter.get_exchangerate_api_response_json(self.currency),
```
arturbalabanov, line 209: Include all the exchange rates currencies **from*** not the **to** currency's exchange rates

app/commands/generate_monthly_charges.py
```diff
@@ -549,6 +530,14 @@ def command(
     exports_dir: Annotated[
         pathlib.Path, typer.Option("--exports-dir", help="Directory to export the charge files to")
     ],
+    max_workers: Annotated[
```
ffaraone, line 533: There's no need in this case to control the number of workers in the thread pool, especially because your code will always use 1 thread.
If you don't pass a thread pool to the `run_in_executor` it will use the default executor (see: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor).

ffaraone (approved):

app/commands/generate_monthly_charges.py
arturbalabanov: Yeah you're right, don't know what I was thinking there :D Changed to the default executor

PR #181, opened 2025-05-06
feat: Add optional account_id and currency filters to generate monthly charges script (MPT-9809)

arturbalabanov:


ffaraone (approved):
There are rooms for optimizations but let's move forward and let's put such optimizations in our tech backlog.

PR #183, opened 2025-05-08
Fix the datasource expense's datasource_id to correctly identify entitlements (MPT-9810)

arturbalabanov:


app/commands/update_current_month_datasource_expenses.py
```diff
@@ -93,13 +93,14 @@ async def store_datasource_expenses(
     for organization_id, datasources in datasources_per_organization_id.items():
         for datasource in datasources:
             existing_datasource_expense, created = await datasource_expense_handler.get_or_create(
-                datasource_id=datasource["id"],
+                linked_datasource_id=datasource["id"],
```
ffaraone, line 96: as a filter for get you should use:

```suggestion
                datasource_id=datasource["account_id"],
```

app/commands/update_current_month_datasource_expenses.py
```diff
@@ -93,13 +93,14 @@ async def store_datasource_expenses(
...
                 organization_id=organization_id,
                 year=year,
                 month=month,
                 defaults={
                     "month_expenses": datasource["details"]["cost"],
                     "datasource_name": datasource["name"],
+                    "datasource_id": datasource["account_id"],
```
ffaraone, line 103: you should add the `linked_datasource_id` to default and, in any case the it should be updated together with the `month_expenses` field.

ffaraone (changes requested):

migrations/versions/0fcca0cf1151_create_datasourceexpense_linked_.py
```diff
@@ -0,0 +1,31 @@
...
+# revision identifiers, used by Alembic.
+revision: str = '0fcca0cf1151'
+down_revision: Union[str, None] = '7cd884694384'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # Add a new column with "" as the default value (so that the migration can be run on a live database)
+    op.add_column('datasource_expenses', sa.Column('linked_datasource_id', sa.String(length=255), server_default="", nullable=False))
```
ffaraone, line 24: This cannot be `nullable=False` because there's data in the table. You can:
1) set initially as `nullable=True`
2) set all the existing records to a blank string
3) set initially as `nullable=False`

ffaraone (changes requested):

app/db/models.py
```diff
@@ -306,8 +310,14 @@ class DatasourceExpense(Base, HumanReadablePKMixin, TimestampMixin):
 
     entitlements: Mapped[list[Entitlement]] = relationship(
         "Entitlement",
-        primaryjoin=lambda: DatasourceExpense.datasource_id == Entitlement.datasource_id,
-        foreign_keys=lambda: [DatasourceExpense.datasource_id],
+        primaryjoin=lambda: (
+            (DatasourceExpense.datasource_id == Entitlement.datasource_id)
```
ffaraone, line 314: Wdyt to add here the condition that the status of the Entitlement must be ACTIVE ?

ffaraone (approved):

app/db/models.py
arturbalabanov: That's a good idea, add that :)

PR #184, opened 2025-05-12
feat: add --dry-run flag to charge files generation command (MPT-9808)

arturbalabanov:


app/commands/generate_monthly_charges.py
```diff
@@ -506,6 +523,7 @@ async def main(
     settings: Settings,
```
ffaraone, line 523: Looks like settings here are not used at all, could we remove the function argument ?

app/commands/generate_monthly_charges.py
```diff
@@ -578,7 +604,13 @@ def command(
         exports_dir.mkdir(parents=True)
 
     asyncio.run(
-        main(exports_dir=exports_dir, currency=currency, account_id=account_id, settings=ctx.obj)
+        main(
+            exports_dir=exports_dir,
```
ffaraone, line 608: Could we default the exports dir to some /tmp/whatever so we can have the options not required ?

ffaraone (approved):

app/commands/generate_monthly_charges.py
arturbalabanov: Removed

app/commands/generate_monthly_charges.py
arturbalabanov: Good idea, implemented this :)

PR #185, opened 2025-05-14
MPT-8302 setup opentelemetry

ffaraone:


app/telemetry.py
```diff
@@ -0,0 +1,69 @@
...
+
+    SQLAlchemyInstrumentor().instrument(engine=dbengine.sync_engine, enable_commenter=True)
+
+
+def capture_telemetry[R](
+    tracer_name: str, span_name: str
+) -> Callable[[Callable[P, Coroutine[Any, Any, R]]], Callable[P, Coroutine[Any, Any, R]]]:
+    def decorator(func: Callable[P, Coroutine[Any, Any, R]]) -> Callable[P, Coroutine[Any, Any, R]]:
+        @wraps(func)
+        async def _wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
```
arturbalabanov, line 61: You can put the `ParamSpec` in the generic defintion by using `capture_telemetry[**P, R]` and remove the `
P = ParamSpec("P")` from above. Or, if you're not a big fan of this syntax, maybe extract R into an explicit `TypeVar`, so that it's more consistent with `ParamSpec`.

I think you can also simplify this a bit by replacing `Coroutine[Any, Any, R]]` with `Awaitable[R]`

---

type annotations for decorators are always fun :D

pyproject.toml
```diff
@@ -44,6 +44,12 @@ dependencies = [
     "rich==14.0.*",
     "pyfiglet==1.0.*",
     "adaptive-cards-py==0.3.*",
+    "opentelemetry-api==1.33.*",
+    "opentelemetry-instrumentation-fastapi==0.54b0",
+    "opentelemetry-instrumentation-logging==0.54b0",
+    "opentelemetry-instrumentation-httpx==0.54b0",
+    "azure-monitor-opentelemetry-exporter==1.0.0b36",
+    "opentelemetry-instrumentation-sqlalchemy==0.54b0",
```
arturbalabanov, line 52: Do we want to pin these pacakges to so specific versions? Though I don't know if these are following semantic releases, so maybe that's why you did it this way...

arturbalabanov (approved):
Looks good, just a couple of nitpicks, feel free to resolve them as you see fit

app/telemetry.py
ffaraone: Thank you!!!

pyproject.toml
ffaraone: Yeah, and we had already experienced breaking changes between "minors".

PR #200, opened 2025-05-21
feat: Opentelementry improvements (MPT-10401)

arturbalabanov:
* A new docker image (Jaeger) added to docker compose for easier testing of local changes (to use it simply use `docker compose up jaeger` and access the UI on `http://0.0.0.0:16686/`)
* Set the `az.namespace` attribute for CLI command spans, so that they are easily identifiable in Azure AppInsight (previously they had `N/A` in the first column)
* Add spans for database transactions allowing us to easily group related queries and hide them when inspecting logs
* Add an optional `--currency` argument to  `update-latest-exchange-rates` CLI command to force the script to fetch the rates for only one currency -- it was useful for testing the OpenTelemetry changes locally and we may want to do that anyway if we want to manually update exchange rates

### DIFF
--- a/app/conf.py
+++ b/app/conf.py
@@ -76,6 +76,7 @@ class Settings(BaseSettings):
     jaeger_host: str = "jaeger"
     jaeger_port: int = 4318
     azure_insights_connection_string: str | None = None
+    opentelemetry_sqlalchemy_min_query_duration_ms: int | None = 100
 
     msteams_notifications_webhook_url: str | None = None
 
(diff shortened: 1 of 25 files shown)


app/conf.py
```diff
@@ -52,6 +52,10 @@ class Settings(BaseSettings):
     azure_sa_max_concurrency: int = 4
     azure_sa_sas_expiration_token_mins: int = 5
 
+    jaeger_enabled: bool = True
```
arturbalabanov, line 55: todo: replace with opentelementry_type

ffaraone (approved):

PR #203, opened 2025-05-30
MPT-10737 Use threading in fetch_datasource_expenses command

bdjilka:


app/conf.py
```diff
@@ -37,6 +37,7 @@ class Settings(BaseSettings):
     datasources_expenses_obsolete_after_months: int = 6
     billing_percentage: float = 1.0
     ffc_external_product_id: str = "FIN-0001-P1M"
+    semaphore_count: int = 10
```
ffaraone, line 40: what about `max_parallel_tasks` ? Sounds clearer to me 😄

app/commands/fetch_datasource_expenses.py
```diff
@@ -80,11 +70,45 @@ async def fetch_datasources_for_organizations(
             organization.id,
             organization.name,
         )
+        filtered_datasources = filter_relevant_datasources(response_datasources)
+        async with lock:
```
ffaraone, line 74: There is no need to lock, there's no other `await` in your task function that can give the loop to some other task, and even in that case, each task handle a given organization so only one task will handle the same dict entry (key).

app/commands/fetch_datasource_expenses.py
```diff
@@ -80,11 +70,45 @@ async def fetch_datasources_for_organizations(
...
+            datasources_per_organization_id[organization.id] = filtered_datasources
+        return
```
ffaraone, line 76: Please remove this redundant `return` so Sonar will be happy 😄

ffaraone (changes requested):
Pretty good as a first step since the most costly operations is the API call, but I would suggest a more radical approach (that can be done into a second iteration), i.e. the task can do either fetch the datasources either insert the records into the db.
Worth (but low prio) to test that we are spawning parallel tasks and the that the task function use a semaphore to limit concurrency.

PR #204, opened 2025-06-05
[MPT-10785] Add GET /v1/expenses endpoint to return a list of expenses

antoniodimariano:
[MPT-10785] Add GET /v1/expenses endpoint to return a list of expenses

[MPT-10785]: https://softwareone.atlassian.net/browse/MPT-10785?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/schemas/expenses.py
+++ b/app/schemas/expenses.py
@@ -1,0 +1,16 @@
+from decimal import Decimal
+
+from app.enums import DatasourceType
+from app.schemas.core import CommonEventsSchema, IdSchema
+from app.schemas.organizations import OrganizationReference
+
+
+class DatasourceExpenseRead(IdSchema, CommonEventsSchema):
+    linked_datasource_id: str
+    datasource_name: str
+    linked_datasource_type: DatasourceType
+    organization: OrganizationReference
+    year: int
+    day: int
+    month: int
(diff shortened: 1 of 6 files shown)


app/schemas/expenses.py
```diff
@@ -0,0 +1,18 @@
...
+class DatasourceExpenseRead(IdSchema, CommonEventsSchema):
+    linked_datasource_id: str
+    datasource_name: str
+    linked_datasource_type: DatasourceType
+    organization: OrganizationReference
+    year: int
+    day: int
+    month: int
+    expenses: Decimal
+    entitlements: list[EntitlementRead]
```
ffaraone, line 18: I think we can remove the list of entitlement furthermore I think we are not fetching them so this attibute will never exist in the response.

ffaraone (changes requested):
Overall perfect except a small glitch, please fix it, good job @antoniodimariano !

app/schemas/expenses.py
antoniodimariano: right

PR #222, opened 2025-07-18
Using SBOM template

plebann:
Replaced the custom implementation of the `dtrack` job with a reusable workflow from `softwareone-platform/ops-template`.
DEPENDENCYTRACK_APIKEY secret is required!


ffaraone (approved):
👍

